PR #11, opened 2016-02-24
Small tweek in seed_from_gml to make it work with matplotlib after 1.2.0

paulskeie:
If you are interested in making seed_from_gml work for newer versions of matplotlib Knut-Frode.


knutfrode:
Takkar!

PR #16, opened 2016-08-08
Moved examples out of the way

vegardb:
Moving example code away from the base directory.


knutfrode:
Several of the examples use data from the folder "test_data", which is also used by the unittests. Thus I moved now all the example scripts to a subfolder "examples", and then updated the paths to the "test_data" folder within those scripts.
https://github.com/knutfrode/opendrift/commit/3b6c57799972f9e3ffc725263b6a151dc85774ca

PR #46, opened 2017-07-17
Added performance updates

babrodtk:
This adds performance improvements to interpolation and basemap coastline checking.

### DIFF
--- a/opendrift/readers/basereader.py
+++ b/opendrift/readers/basereader.py
@@ -76,7 +76,7 @@ class BaseReader(object):
                          # capable of returning blocks of data
 
     # Default interpolation method, see function interpolate_block()
-    interpolation = 'ndimage'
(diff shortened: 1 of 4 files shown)


babrodtk:
Appears that the reference solutions have changed because of the changed default interpolation method. It should be possible to leave the linear (existing default).

opendrift/readers/basereader.py
```diff
@@ -76,7 +76,7 @@ class BaseReader(object):
                          # capable of returning blocks of data
 
     # Default interpolation method, see function interpolate_block()
-    interpolation = 'ndimage'
+    interpolation = 'linearNDFast'
```
babrodtk: This changes the default interpolation from nearest neighbor to

knutfrode:
Merged manually.

PR #48, opened 2017-07-28
Parameterisation of stokes drift and significant wave height if not a…

petternygren:
Simple methods for selecting pre calculated stokes drift velocity and significant wave height based on wind speed and fetch. Tables range from 0 to 30 m/s. Fetch 5000m, 25000m and 50000m.

Just a nice feature to have if working with limited input. The feature was added to opendrift in order to easier compare results with the SMHI driftmodel SeaTrackWeb.

**Parameterised with fetch 25000m:**
animate_25000m_fetch

**No parameterisation**
animate_no_parameterisation

### DIFF
--- a/opendrift/models/basemodel.py
+++ b/opendrift/models/basemodel.py
@@ -619,7 +619,7 @@ class OpenDriftSimulation(PhysicsMethods):
 
         return variable_groups, reader_groups, missing_variables
 
-    def get_environment(self, variables, time, lon, lat, z, profiles):
+    def get_environment(self, variables, time, lon, lat, z, profiles, parameterise=True):
         '''Retrieve environmental variables at requested positions.
 
         Updates:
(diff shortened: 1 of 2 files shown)


knutfrode:
Thank you Petter, this is a very useful feature.

It would be good to do a small change, to be compliant with the more general usage:
The ```parameterise=True``` parameter of get_environment() is not ideal, as this method is used also for modules which don't use or care about Stokes drift (e.g. the WindDrift module). This it would be better to rather control this with the recently introduced config mechanism, e.g. something like:
```
o.set_config('drift:use_tabularised_stokes_drift', True)
```
from the users side, and 
```
self.get_config('drift:use_tabularised_stokes_drift')
```
from within OpenDrift.

I will have a closer look at this when I am back from holiday 10th August.

petternygren:
Yes I'll add turning the feature on/off via the config. Together with the option too choose fetch distance.

petternygren:
Ok, this feature can now be activate via two config items.

drift:use_tabularised_stokes_drift = boolean(default=False)
drift:tabularised_stokes_drift_fetch = option(5000, 25000, 50000, default=25000)

PR #66, opened 2018-03-12
Addition of a new reader and module SedimentDrift3D to codebase

simon-msl:
Hi Knut - here a first pull request. Key changes summarised below (most in my previous email but reproducing here for reference):

- No edits to the existing code base ( and I made sure to stay up to date with your commits as I was developing).

- New "reader" to load netcdf files formatted according to internal convention (which is CF in some parts but not always). We are in the process of releasing an range of  open-datasets for NZ so ideally people could use that reader to load that data and play with it. Added some code to allow logarithmic extrapolation of depth-averaged currents to any vertical level in water column.

- New "model" called sedimentdrift3d.py which includes the class SedimentDrift3D. I used this module to experiment with things that may be relevant elsewhere e.g. horizontal diffusion coefficient, resuspension. This includes a modified get_environment() to allow adding currents from different readers rather that choosing one of the other - probably need to find a more elegant way to do this to allow adding tide+residual, while also retaining the dataset priority feature. An option may be to use [*_sea_water_velocity_tidal, *_sea_water_velocity_geostrophic, *_sea_water_velocity (i.e. total)] for the _required_ modules variables (will need to update the variable mapping accordingly), and assess whether we need to combine currents (e.g. if tidal+residuals) or not (total current or single dataset) in the _get_environement()_

- Some initial work on the "docker"-ization of the code. The idea is to be able to have a docker image that has opendrift and all other third-party libraries installed  (OilLibrary from NOAA, Motu-Client from Mercator). The idea would be to eventually push an image to dockerhub so that people can simply 'pull' it, with one line of code, and start using the full model .

Iam relatively new to the github process - and I realized that you may want to discard the example_msl folder, but not too sure how to keep that in github (for me) and not include it in that pull request ...

Feel free to suggest a better way to collaborate on this if there is one. 

Best regards,
Simon


ADevisM:
hello.
I have downloaded the OpenDrift 1.0,.4 code, but I could not find  the module SedimentDrift3D or any specific way to simulate sediment. I have used before the OpenDrift3D before but now I need to simulate sediments. Help please.

PR #130, opened 2019-09-06
python3: issues encountered during setup

gauteh:
works with oillibrary py 3 port (https://github.com/NOAA-ORR-ERD/OilLibrary/pull/18)


conda_python3.yml
```diff
@@ -1,19 +1,29 @@
...
   - ipython
+  - pip
+  - pytest
+  - GitPython
+  - SQLAlchemy>=1.3
+  - zope.sqlalchemy>=1.1
+  - awesome-slugify
+  - unit_conversion>=2.5
+  - pip:
+    - https://github.com/matplotlib/basemap/archive/a551c2361314670dc7d95143190d3d48ba750d20.zip
```
gauteh: Check if better version?

opendrift/models/basemodel.py
```diff
@@ -521,7 +523,7 @@ def set_projection(self, proj4):
     def lonlat2xy(self, lon, lat):
         """Calculate x,y in own projection from given lon,lat (scalars/arrays).
         """
-        if self.proj.is_latlong():
+        if self.proj.crs.is_geographic:
```
gauteh: Wont work on old pyproj versions (<2)

conda_python3.yml
```diff
@@ -1,19 +1,29 @@
...
+    - git+https://github.com/gauteh/OilLibrary.git@python3
```
gauteh: point to opendrift fork

opendrift/models/basemodel.py
gauteh: pyproj upgraded for py 2

gauteh:
Updated to support python 2 as well. Tests currently fail since travis deps are not synced with conda environment files.

gauteh:
The reason for the error in travis was related to basemap 1.2.1 not supporting pyproj >= 2. I've set up to build basemap from master since there are no new releases with support for pyproj (yet). I think the ultimate fix to this is moving to cartopy (#135).

PR #145, opened 2019-10-03
Cartopy reader

gauteh:


knutfrode:
example.py crashes with error
```
  File "/home/knutfd/software/opendrift/opendrift/readers/reader_cartopy_landmask.py", line 75, in __init__
    cachef = old.GSHHS[resmap[resolution]]
AttributeError: 'module' object has no attribute 'GSHHS'
```

If I change `GSHHS` to `GSSHS`in reader_cartopy_landmask.py, I am getting this error:
```
IOError: [Errno 2] No such file or directory: '/home/knutfd/miniconda2/envs/opendrift_p2_oil/lib/python2.7/site-packages/opendrift_landmask_data/shapes/gshhs_h_-180.000000E-90.000000N180.000000E90.000000N.wkb'
```

gauteh:
Try uninstalling and re-installing `opendrift-landmask-data`, I fixed it in the most recent pypi version. Install using normal pip: `pip install opendrift-landmask-data`.

gauteh:
It is probably still going to fail, seems that the files are not included in the py2 version of the package yet.

gauteh:
Uploaded a new version of the landmask-data.

knutfrode:
Now it works well!
example.py now initialises the landmask much faster than with basemap (0.1s versus 13s). 

However, during the simulation, cartopy landmask uses 2.7 seconds versus 0.7 seconds for basemap. 
Increasing the number of elements from 3000 to 300.000, the cartopy landmask uses 2min48seconds, versus 19.7 seconds for basemap landmask.

This is probably due to the rasterisation implemented in the basemap reader. This machanism pre-screens most of the elements, so that only a small subset needs to be checked for land.
I guess it is straightforward to transfer this mechanism to the cartopy reader.

gauteh:
Yes, we can maybe use something like https://github.com/mapbox/rasterio or perhaps shapely already has something native which we can use to pre-generate the raster-images. As long as they don't end up being gigantic. Anyway with a good choice of memory layout (maybe it has to be segmented into areas and not x,y entire world) it should benefit greatly from memory-mapping since most points will be relatively localized and we therefore only need to read a subset of the file. 

The way it is done in basemap reader now we need to make sure that the rasterization is inclusive, so that we never get false negatives, then we should be safe when using coarse resolution rasterization and high resolution polygons. This probably best since relatively few points compared to the number of active points hit land and need to be further checked in each iteration.

knutfrode:
Pre-generated rasters might work well for lower resolutions, but probably not feasible for highest resolutions, where we would need pixels on the order of a km or less.
Thus I was thinking of dynamical rasterisation at startup, which means that we would again need the extent (given by user or estimated at run startup, as presently).
This consumes some time at startup, but quickly pays back for longer (realistic, non-unittest) simulations or many particles. The present pixelisation is secure, it is just a pre-screening, and all particles within a land-pixel are subsequently checked against polygons.

gauteh:
Another alternative is using RTree's [0][1]. As far as I understand this means that we split the polygons into tiles, this would create far less polygons to check each point against since we can determine explicitly which tile the point belongs in. Since we need to check against polygons at a later time this will avoid the second step.

The tiled shapefiles can in theory be stored in the same way we do now, and should not cause any additional startup-cost.

[0] https://stackoverflow.com/a/36299183/377927
[1] https://gis.stackexchange.com/questions/120955/understanding-use-of-spatial-indexes-with-rtree/144764#144764

gauteh:
We can also detect empty (ocean-only) tiles, and avoid checking those (not that it would take any time, since there are no polygons to check points against). Then we approach rasterization with polygons on lower level.

knutfrode:
Yes, this sounds very promising. This also removes the present danger of running out of estimated extent.

gauteh:
Merged.

PR #146, opened 2019-10-08
Densitymapping

johannesro:
- allow to compute density maps for any given map projection or given region 
- use mollweide projection as default for density maps
- new write_netcdf_density_map_proj() function to use these features
- write_netcdf_density_map() imitates old longlat grid interface

### DIFF
--- a/opendrift/models/basemodel.py
+++ b/opendrift/models/basemodel.py
@@ -2598,8 +2598,9 @@ class OpenDriftSimulation(PhysicsMethods):
                   background=None, vmin=None, vmax=None, drifter=None,
                   skip=5, scale=10, color=False, clabel=None,
                   colorbar=True, cmap=None, density=False, show_elements=True,
-                  show_trajectories=False, density_proj=None,
-                  llcrnrlon=None, llcrnrlat=None, urcrnrlon=None, urcrnrlat=None,
+                  show_trajectories=False, 
+                  #density_proj=None,
+                  #llcrnrlon=None, llcrnrlat=None, urcrnrlon=None, urcrnrlat=None,
                   density_pixelsize_m=1000, unitfactor=1, lcs=None,
                   surface_only=False, markersize=20,
                   legend=None, legend_loc='best', fps=10):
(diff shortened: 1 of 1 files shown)


opendrift/models/basemodel.py
```diff
@@ -2598,7 +2598,9 @@ def animation(self, buffer=.2, corners=None, filename=None, compare=None,
                   background=None, vmin=None, vmax=None, drifter=None,
                   skip=5, scale=10, color=False, clabel=None,
                   colorbar=True, cmap=None, density=False, show_elements=True,
-                  show_trajectories=False,
+                  show_trajectories=False, 
+                  #density_proj=None,
+                  #llcrnrlon=None, llcrnrlat=None, urcrnrlon=None, urcrnrlat=None,
```
knutfrode: Debugging lines should be removed

opendrift/models/basemodel.py
```diff
@@ -2780,9 +2782,12 @@ def plot_timestep(i):
             cmap.set_under('w')
             H, H_submerged, H_stranded, lon_array, lat_array = \
                 self.get_density_array(pixelsize_m=density_pixelsize_m,
+                                       #density_proj=density_proj,
+                                       #llcrnrlon=llcrnrlon, llcrnrlat=llcrnrlat,
+                                       #urcrnrlon=urcrnrlon, urcrnrlat=urcrnrlat,
```
knutfrode: Debugging lines should be removed

opendrift/models/basemodel.py
```diff
@@ -3294,39 +3299,52 @@ def get_map_background(self, map, background, time=None):
...
-        lat_array = np.arange(np.nanmin(lat)-deltalat,
-                              np.nanmax(lat)+deltalat, deltalat)
-        lon_array = np.arange(np.nanmin(lon)-deltalat,
-                              np.nanmax(lon)+deltalon, deltalon)
-        bins=(lon_array, lat_array)
+        if density_proj is None: # add default projection with equal-area property 
+            density_proj = pyproj.Proj('+proj=moll +ellps=WGS84 +lon_0=0.0')
+
+        # create a grid in the specified projection
+        x,y = density_proj(lon, lat)
```
knutfrode: I discovered that the new write_netcdf_density_map() writes a grid which is much too small (e.g. 3x4 instead of 150x200 pixels). The reason (or part of the reason) seems to be that density_proj(lon, lat) returns not x,y in meters, but lon, lat, and apparently in radians! This might be an explanation:
http://pyproj4.github.io/pyproj/stable/gotchas.html#proj-not-a-generic-latitude-longitude-to-projection-converter

knutfrode:
Fixed in #523, but by duplicating two methods. These have to be merged later.

PR #147, opened 2019-10-09
Radionuclide

magnesim:


opendrift/models/radionuclides.py
```diff
@@ -0,0 +1,620 @@
...
+#
+# OpenDrift is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with OpenDrift.  If not, see <http://www.gnu.org/licenses/>.
+#
+# Copyright 2015, Knut-Frode Dagestad, MET Norway
```
knutfrode: You can here use your own name!

opendrift/models/radionuclides.py
```diff
@@ -0,0 +1,620 @@
...
+
+import numpy as np
+
+from opendrift.models.opendrift3D import \
+    OpenDrift3DSimulation, Lagrangian3DArray
+from opendrift.elements import LagrangianArray
+import logging
+from __builtin__ import True
```
gauteh: Is this line necessary?

opendrift/models/radionuclides.py
```diff
@@ -0,0 +1,620 @@
...
+        
+
+        
+    
+    
+        
+
+
+    def prepare_run(self):
+        print 'Preparing'
```
gauteh: Won't work on python3

opendrift/models/radionuclides.py
```diff
@@ -0,0 +1,620 @@
...
+        if self.get_config('radionuclide:species:Sediment_slowly_reversible') and \
+                    self.get_config('radionuclide:species:Particle_slowly_reversible'):
+            self.set_config('radionuclide:slowly_fraction', True)
+        if self.get_config('radionuclide:species:Sediment_irreversible') and \
+                    self.get_config('radionuclide:species:Particle_irreversible'):
+            self.set_config('radionuclide:irreversible_fraction', True)
+        
+        
+        self.nspecies      = len(self.name_species)
+        print 'Number of species: {}'.format(self.nspecies)
```
gauteh: Python3

opendrift/models/radionuclides.py
```diff
@@ -0,0 +1,620 @@
...
+            self.set_config('radionuclide:slowly_fraction', True)
+        if self.get_config('radionuclide:species:Sediment_irreversible') and \
+                    self.get_config('radionuclide:species:Particle_irreversible'):
+            self.set_config('radionuclide:irreversible_fraction', True)
+        
+        
+        self.nspecies      = len(self.name_species)
+        print 'Number of species: {}'.format(self.nspecies)        
+        for i,sp in enumerate(self.name_species):
+            print '{:>3} {}'.format( i, sp ) #, self.specie_name2num(sp), self.specie_num2name(i)
```
gauteh: Same, any print without `()`, skipping the rest of those :)

opendrift/models/radionuclides.py
```diff
@@ -0,0 +1,620 @@
...
+
+import numpy as np
+
+from opendrift.models.opendrift3D import \
+    OpenDrift3DSimulation, Lagrangian3DArray
+from opendrift.elements import LagrangianArray
+import logging
+from __builtin__ import True
+from numpy.distutils.fcompiler import none
+from ansible.constants import get_config
```
gauteh: Is this a new dependency?

opendrift/models/radionuclides.py
```diff
@@ -0,0 +1,620 @@
...
+        self.elements.diameter[(sp_out==self.num_lmm) & (sp_in!=self.num_lmm)] = dia_diss
+        
+        # Transfer to colloids
+        if self.get_config('radionuclide:species:Colloid'):
+            self.elements.diameter[(sp_out==self.num_col) & (sp_in!=self.num_col)] = dia_diss
+
+
+
+
+    def bottom_interaction(self,Zmin=none):
```
gauteh: This seems like the only place where `none` (lowercase) is used, was that intended? and if not then one import can also be removed.

opendrift/models/radionuclides.py
knutfrode: I suggest to replace the print-line with:
```logging.debug('Preparing radionucleide species')```

File: opendrift/models/radionuclides.py [hunk 8]
```diff
@@ -0,0 +1,620 @@
...
+        logging.info('Transfer rates: %s' % self.transfer_rates)
+
+
+
+
+
+
+        
+
+    def update_terminal_velocity(self, Tprofiles=None,
```
knutfrode: This method seems to be copied directly from pelagicegg.py. It unfortunate to duplicate code, so ideally it should be either imported from pelagicegg.py, or both these modules should import it from another module/class. But this can also be done later.

magnesim: Indeed, it is copied. However it is slightly modified, using particle density from a preset element property, instead of computing it. I agree that it is not fortunate to have duplicates.

gauteh:
Strange, did you @magnesim include the changes from the rebase?

ons. 9. okt. 2019, 16:36 skrev Knut-Frode Dagestad <notifications@github.com

gauteh:
Skjermbilde 2019-10-09 kl  16 59 50

magnesim:
Well, not intentionally, as far as I know...

knutfrode:
Screenshot 2019-10-09 at 17 03 21

magnesim:
But, as you may have noticed, I just pushed a new commit. With some clean-up in the radionuclide module and the example script

gauteh:
Weird, maybe we found a github bug :)

PR #152, opened 2019-10-16
Use cartopy for plotting

gauteh:
(Currently depends on #145, but does not need to)

Not all old functionality has been moved yet.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ script:
       docker exec "opendrift_${TRAVIS_COMMIT}" bash -c ". ~/.bashrc && conda install -q --yes -c conda-forge xarray=$XARRAY dask";
     fi
   - docker exec "opendrift_${TRAVIS_COMMIT}" bash -c ". ~/.bashrc && conda list"
-  - docker exec "opendrift_${TRAVIS_COMMIT}" bash -c ". ~/.bashrc && pytest --cov=opendrift --ignore=tests/benchmarks --ignore=tests/wps --runslow --durations=10"
+  - docker exec "opendrift_${TRAVIS_COMMIT}" bash -c ". ~/.bashrc && pytest --cov=opendrift --ignore=tests/benchmarks --ignore=tests/wps --runslow --durations=10 --benchmark-disable"
 
 after_success:
   - docker cp "opendrift_${TRAVIS_COMMIT}":/code/.coverage .
--- a/examples/example.py
+++ b/examples/example.py
@@ -45,9 +45,9 @@ o.run(end_time=reader_norkyst.end_time, time_step=1800,
 # Print and plot results
 print(o)
 #o.plot(background=['x_sea_water_velocity', 'y_sea_water_velocity'], buffer=.5)
-o.animation()
-o.animation(density=True, show_elements=False)
+o.animation(fast=True)
+o.animation(density=True, show_elements=False, fast=True)
 #o.animation(filename='openoil_time_seed.gif')
-o.plot()
+o.plot(fast=True)
 #o.plot_property('mass_oil')
 #o.plot_property('x_sea_water_velocity')
--- a/examples/example_LCS_norkyst.py
+++ b/examples/example_LCS_norkyst.py
@@ -6,7 +6,6 @@ from opendrift.readers import reader_netCDF_CF_generic
 from opendrift.models.oceandrift import OceanDrift
 
 o = OceanDrift(loglevel=20)  # Set loglevel to 0 for debug information
-o.set_config('general:basemap_resolution', 'i')
 
 # Norkyst ocean model
 reader_norkyst = reader_netCDF_CF_generic.Reader(o.test_data_folder() + '16Nov2015_NorKyst_z_surface/norkyst800_subset_16Nov2015.nc')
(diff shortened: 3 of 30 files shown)


knutfrode:
After checkout and updating the conda3-environment, I am getting this error when trying to import the landmask reader:
```
  File "/home/knutfd/software/opendrift/opendrift/readers/reader_global_landmask.py", line 18, in <module>
    from opendrift_landmask_data import Landmask
  File "/home/knutfd/miniconda2/envs/opendrift_p3/lib/python3.7/site-packages/opendrift_landmask_data/__init__.py", line 2, in <module>
    from .mask import Landmask
  File "/home/knutfd/miniconda2/envs/opendrift_p3/lib/python3.7/site-packages/opendrift_landmask_data/mask.py", line 6, in <module>
    import rasterio
  File "/home/knutfd/miniconda2/envs/opendrift_p3/lib/python3.7/site-packages/rasterio/__init__.py", line 22, in <module>
    from rasterio._base import gdal_version
ImportError: libzstd.so.1: cannot open shared object file: No such file or directory
>>> 
```

knutfrode:
Deleting and recreating the environment resolved the above issue, so seems to be a conda issue.

First run of an example script worked well, and spent ~30 seconds for the rasterisations, but startup was immediate at second run.
The plotting of map at the end, was however very slow, as it first needed to download polygons. But should these polygons not already be locally available? Also zooming on a figure takes more time than with Basemap, for some reason?
I will do some more testing later.

gauteh:
> First run of an example script worked well, and spent ~30 seconds for the rasterisations, but startup was immediate at second run.

Good, the rasterization has already been done and this created a memory-map file in `/tmp`, which you can delete if you want and it will be re-created automatically.

> The plotting of map at the end, was however very slow, as it first needed to download polygons. But should these polygons not already be locally available? 

The polygons (full resolution) are only downloaded for the mask (in pre-processed form). The landmask reader is independent of cartopy and should be fully functional without cartopy installed. These are not re-used since we would then deviate from standard use of cartopy and would have to re-implement the intersection / polygon selection functionality, as well as miss any updates that might happen upstream on cartopy here. Additionally, we only include 'full' resolution which is too slow for most quick iterations.

> Also zooming on a figure takes more time than with Basemap, for some reason?
> I will do some more testing later.

Maybe it works faster if we use the same projection as the shapes (PlateCarree). It might otherwise be worth testing out using one of the tiling services as well, there are many options included by default (including [satellite images](https://scitools.org.uk/cartopy/docs/latest/gallery/wmts_time.html#sphx-glr-gallery-wmts-time-py) and google maps sources). Images should be faster.

gauteh:
Unfortunately, Python 3 appears to be [slower than Python 2](https://stackoverflow.com/questions/23453133/is-there-a-reason-python-3-enumerates-slower-than-python-2) for a bunch of cases. I noticed this with the landmask checker which is almost 3x slower for the polygon checker on python3 compared to python2 (not the mask check - but the polygon check). I don't know if there is a good way to work-around this. Forcing  `dtype = np.int32` does not help.

knutfrode:
Yes, the testall script takes now 436 seconds with Python3 and cartopy_plot on my laptop, whereas this used to take about 90 seconds on Python2 and basemap. So we need to detect and resolve the performance bottlenecks. I will have another look tomorrow.

gauteh:
> Yes, the testall script takes now 436 seconds with Python3 and cartopy_plot on my laptop, whereas this used to take about 90 seconds on Python2 and basemap. So we need to detect and resolve the performance bottlenecks. I will have another look tomorrow.

There are also more tests now which probably affect that, and also a bunch of benchmark tests which should not be included in the default test runs, but have unit-test equivalents.

gauteh:
Removed some duplicate tests and only run benchmarks once by default:

`$ ./testall --durations=20`

```
======================================================================== slowest 20 test durations ========================================================================
50.31s call     tests/integration/test_run.py::TestRun::test_plot_animation
17.34s call     tests/integration/test_run.py::TestRun::test_reader_order
7.50s call     tests/integration/test_run.py::TestRun::test_export_step_interval
7.41s call     tests/integration/test_run.py::TestRun::test_output_time_step
6.01s call     tests/integration/test_stranding.py::TestStranding::test_stranding_roms
5.37s call     tests/integration/test_stranding.py::TestStranding::test_stranding_options
5.15s call     tests/integration/test_stranding.py::TestStranding::test_interact_coastline_global
4.60s call     tests/integration/test_models.py::TestModels::test_shipdrift_backwards
4.60s call     tests/integration/test_models.py::TestModels::test_shipdrift
4.53s call     tests/unit/test_global_landmask.py::test_global_array
4.49s call     tests/integration/test_physics.py::TestPhysics::test_vertical_mixing_plantoil_windonly
3.90s call     tests/unit/test_global_landmask_basemap.py::test_basemap_setup
3.84s call     tests/integration/test_readers.py::TestReaders::test_repeated_run
3.13s call     tests/integration/test_stranding.py::TestStranding::test_stranding_3d
2.96s call     tests/integration/test_readers.py::TestReaders::test_lazy_reader_oildrift
2.90s call     tests/integration/test_oillibrary.py::TestOil::test_dispersion
2.68s call     tests/integration/test_readers.py::TestReaders::test_constant_reader
2.64s call     tests/integration/test_netcdf.py::Test::test_MFDataset
2.60s call     tests/integration/test_readers.py::TestReaders::test_clip_domain
2.41s call     tests/unit/test_global_landmask.py::test_landmask_global
======================================================== 103 passed, 3 skipped, 118 warnings in 189.51s (0:03:09) =========================================================
```

PR #155, opened 2019-10-28
setup.py for the new millennium 

gauteh:
Getting ready for pypi and conda releases of opendrift.


gauteh:
Now based on cartopy_plot

PR #156, opened 2019-10-31
Small fix for FTLE calculation

knutfrode:


examples/example_codegg.py
```diff
@@ -44,7 +44,6 @@
 # Print and plot results
 print(o)
 
-o.plot(filename='codegg_forecast.png')
-o.animation(filename='codegg_forecast.gif')
-o.plot_vertical_distribution(filename='codegg_vertical.png')
-
+o.plot(filename='codegg_forecast.png', fast=True)
+o.animation(filename='codegg_forecast.mp4', fast=True)
+o.plot_vertical_distribution()
```
gauteh: Was filename intentionally left out here?

gauteh:
One comment, otherwise looks good :+1:

examples/example_codegg.py
knutfrode: Yes, this plotting method is interactive with slider, and does not support saving to file.

PR #159, opened 2019-11-01
Now using a common, named logger (opendrift) for all modules. syntax …

knutfrode:
…is now self.logger.deug() instead of logging.debug()


gauteh:
Looks good 👍

PR #160, opened 2019-11-01
Documentation using sphinx and readthedocs

gauteh:
http://opendrift.github.io


gauteh:
Building docs on readthedocs.org with opendrift simulations takes too long and too much memory for readthedocs and is probably not possible. It is possible to build as part of circleci job, deploy to a separate doc repository and serve at e.g.: http://opendrift.github.io (or arbitrary domain).

gauteh:
See https://opendrift.github.io/models.html for the end-result with automatically generated opendrift figures. Currently this shows the last built branch, so we should obviously update that to only master (but try to build docs without publishing on all). Alternatively use something like https://sphinxcontrib-versioning.readthedocs.io/en/latest/tutorial.html to have multiple versions.

We should also remove the readthedocs.org version.

gauteh:
Yes, I agree

lør. 2. nov. 2019, 11:25 skrev Knut-Frode Dagestad <notifications@github.com

gauteh:
I've updated the conda_python3 environment as well as instructions for setting up for installing from source. I think this method is best way for developers to use opendrift. All pip requirements are gone (with the new opendrift conda repository), so it is now a pure conda solution.

gauteh:
I also tested on Mac OS X. The only way to make this work reliably is to use conda, but that works pretty well. 

Maybe it makes sense to merge this PR soon, if we are sure we want to use this approach for documentation?

knutfrode:
Yes, I agree to merge - documentation looks very good, almost professional.

gauteh:
Hehe, almost professional is pretty good.

gauteh:
I set the homepage of opendrift to the new docs. Feel free to remove if you do not think it is ready yet.

knutfrode:
We should update Readme also. Probably just the same link there, and moving also the Docker part to readthedocs.

gauteh:
Yes, I agree.

ons. 6. nov. 2019 kl. 10:08 skrev Knut-Frode Dagestad <
notifications@github.com>:

PR #163, opened 2019-11-05
Radionuclide

magnesim:


examples/example_radionuclides.py
```diff
@@ -19,89 +19,140 @@
...
-#init_speciation = np.random.randint(o.nspecies,size=ntraj)
-init_speciation = np.ones(ntraj)*0
-
-diam=np.zeros_like(init_speciation,dtype=np.float32)
-#diam[init_speciation==o.name_species.index('Particle reversible')] = 0.001
-#diam[init_speciation==o.name_species.index('Particle irreversible')] = 0.001
-#print 'diam',diam
 
+# Sandnesfjorden 2 case
+reader_sandnesfj = reader_ROMS_native.Reader('/home/magnes/projects/openDrift-RAD/data/norfjords_32m_his.nc4_20180???00-2018????23')
```
knutfrode: This file is only available on Magne's machine, and example will not work for other people. This is not a problem now during development, but eventually it would be good that examples work for all users. However, it is probably not so easy to make nice examples based on the small/space example ocean datasets included.

opendrift/models/radionuclides.py
```diff
@@ -452,23 +551,37 @@ def update_speciation(self):
         # Set the new speciation
         self.elements.specie=specie_out
         
-        logging.debug('old species: %s' % specie_in[phaseshift])
-        logging.debug('new species: %s' % specie_out[phaseshift])
+        self.logger.debug('old species: %s' % specie_in[phaseshift])
+        self.logger.debug('new species: %s' % specie_out[phaseshift])
+
+
+        for iin in xrange(self.nspecies):
```
knutfrode: For Python3, ```xrange``` does not exist, and one has to use ```range``` instead. It is recommended to start using Python3, as Python2 will not be supported much longer.

magnesim:
I have now fixed the use of 'xrange', and submitted a more general example script.

PR #166, opened 2019-11-07
grb reader: py3 + test load

gauteh:


gauteh:
Tested by lars robert.

PR #231, opened 2020-01-28
Update eddydiffusivity.py

simonweppe:
This was needed..at least in my working version. Thanks.


knutfrode:
You are right. A bug seems to have sneaked in sometime.
Thank you!

PR #234, opened 2020-02-03
back-extent

gauteh:


opendrift/readers/basereader.py
```diff
@@ -1284,11 +1285,11 @@ def get_timeseries_at_position(self, lon, lat, variables=None,
             if end_time is None:
                 end_time = self.end_time
             times = [t for t in self.times if t >= start_time and t<= end_time]
-        
+
         if variables is None:
             variables = self.variables
 
-        if len(self.covers_positions(lon=lon, lat=lat)[0]) is 0:
```
knutfrode: This change gives the following warning:
```
SyntaxWarning: "is" with a literal. Did you mean "=="?
  if len(self.covers_positions(lon=lon, lat=lat)[0]) is 0:
```

PR #241, opened 2020-02-06
hide_landmask also available for basemodel.animation(). Improved get_…

knutfrode:
…map_background(), but more improvement is needed here.


knutfrode:
Eg la til ```hide_map_background``` i ```animation()```, siden funksjonaliteten eller kun var tilgjengeleg for ```plot()```

knutfrode:
Ignoring that Conda build failed temporarily on Travis. All other tests are passing.

PR #249, opened 2020-02-26
If ocean_vertical_diffusivity is not provided by ocean model, paramet…

knutfrode:
…erisation from Large1994 will now be used instead of fallback value of 0.02. Tests are updated since mixing is now slightly different.

### DIFF
--- a/opendrift/models/opendrift3D.py
+++ b/opendrift/models/opendrift3D.py
@@ -334,7 +334,7 @@ class OpenDrift3DSimulation(OpenDriftSimulation):
  
         self.timer_end('main loop:updating elements:vertical mixing')
 
-    def plot_vertical_distribution(self, maxdepth=None, numtimes=5):
+    def plot_vertical_distribution_new(self, maxdepth=None, numtimes=5):
         """Function to plot vertical distribution of particles"""
         import matplotlib.pyplot as plt
 
@@ -366,7 +366,7 @@ class OpenDrift3DSimulation(OpenDriftSimulation):
         plt.legend()
         plt.show()
 
-    def plot_vertical_distribution_slider(self):
+    def plot_vertical_distribution(self):
         """Function to plot vertical distribution of particles"""
         import matplotlib.pyplot as plt
         from matplotlib.widgets import Slider, Button, RadioButtons
--- a/opendrift/models/opendrift3D.py
+++ b/opendrift/models/opendrift3D.py
@@ -113,12 +113,10 @@ class OpenDrift3DSimulation(OpenDriftSimulation):
         # do nothing 
         pass
 
-    def get_diffusivity_profile(self):
-        Nparticles = self.num_elements_active()
+    def get_diffusivity_profile(self, model):
         depths = self.environment_profiles['z']
         wind, depth = np.meshgrid(self.wind_speed(), depths)
 
-        model = self.get_config('turbulentmixing:diffusivitymodel')
         if model == 'windspeed_Large1994':
             return verticaldiffusivity_Large1994(wind, depth)
         elif model == 'windspeed_Sundby1983':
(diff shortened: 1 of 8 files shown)


opendrift/models/opendrift3D.py
```diff
@@ -338,6 +334,37 @@ def vertical_mixing(self):
...
+        ax.set_xlabel('number of particles')
+        ax.set_ylabel('depth [m]')
+        #x_wind = self.history['x_wind'].T[tindex, :]
+        #y_wind = self.history['y_wind'].T[tindex, :]
+        #windspeed = np.mean(np.sqrt(x_wind**2 + y_wind**2))
+        #mainplot.set_title(str(self.get_time_array()[0][tindex]) +
+        #                   #'   Percent at surface: %.1f %' % percent_at_surface)
+        #                   '   Mean windspeed: %.1f m/s' % windspeed)
+        plt.legend()
+        plt.show()
```
gauteh: Would this show be problematic on non-desktop?

knutfrode: Yes, possibly. Therefore I removed this new method in the previous commit.
On the other hand, I noticed that also "example_long...." scripts are now run on Circle-CI.
So seems like the filtering of these has disappeared? I guess that would explain the problems?

gauteh: hm, nothing has changed with that that I can see. They will be listed and documented, but not run.

gauteh:
Could try to increase the no output timeout: https://circleci.com/docs/2.0/configuration-reference/#run on the make html step

opendrift/models/opendrift3D.py
knutfrode: ok, I thought the listing meant that they were also run.
But then lets hope the last commit will solve the issue.

gauteh: I increased the timeout. And also the verbosity of example scripts run. This is on master, and currently building. You could rebase this PR and push again to get those changes, but might be worth waiting for the current run to finish first..?

gauteh:
Seeing lots of these errors in the output on master:

```
19:03:13 INFO: Exception:
19:03:13 INFO: '<' not supported between instances of 'list' and 'int'
19:03:13 DEBUG: Traceback (most recent call last):
  File "/root/project/*********/models/basemodel.py", line 947, in get_environment
  File "/root/project/*********/readers/basereader.py", line 530, in get_variables_interpolated
    self._get_variables(blockvariables_before, profiles,
  File "/root/project/*********/readers/basereader.py", line 380, in _get_variables
    (env[variable] < standard_names[variable]['valid_min']) |
TypeError: '<' not supported between instances of 'list' and 'int'
```

gauteh:
https://circleci.com/gh/OpenDrift/opendrift/1471?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

gauteh:
Looks like this was introduced with the valid_min to valid_max checks from 1f5b30e, I don't immediately see a fix.

knutfrode:
> Looks like this was introduced with the valid_min to valid_max checks from [1f5b30e](https://github.com/OpenDrift/opendrift/commit/1f5b30e80ca7ba6ab0b5c08cb5f940b3d23f5163), I don't immediately see a fix.

Yes, and this is not covered by any test, but might happen with online data. I might try to fix with:
```
invalid = np.where(
                    (np.array(env[variable]) < np.float(standard_names[variable]['valid_min'])) |
                    (np.array(env[variable]) > np.float(standard_names[variable]['valid_max'])))[0]
```

gauteh:
Ok, sure you should not use `logical_or` (https://docs.scipy.org/doc/numpy/reference/generated/numpy.logical_or.html) bitwise with numpy arrays does not always work as expected.

knutfrode:
A bit strange that disabling example_interpolation solved the problem, as this one is short and runs quickly?

gauteh:
I think that the reason the scripts timed out was the bug we found above. I tried rebasing to get the verbose version (so that you can see the error on circleci) and taking in your changes, except disabling any scripts in #250.

gauteh:
#250 gjekk igjennom (samme som denne, men rebasa og med ingen deaktiverte eksempel). Men det er noko som går treigare, det tok over 2 timer å køyre testane. Før tok det rundt 40 min.

knutfrode:
Last commit failed after 34 minutes (10min with no output), but completed successfully second time (git amend) after 1hr14min. Thus the problem should be variable CircleCI slowness, and not the code.
I guess we can then merge this PR.

gauteh:
I think the way to know for sure is to rebase this PR on top of master to get: _higher verbosity from circleci_ and _longer timeout on no output_. I think that the commits disabling some examples can be removed since the problem was not the examples, but the other bug. To to this:
```
$ git checkout master
$ git pull
$ git checkout dev
$ git rebase -i master
# remove lines with commits disabling examples
$ git push -f # if everything went well
```

you can `git checkout -b dev-backup; git checkout dev` from `dev` branch first to make a backup of the branch before starting. `git rebase --abort` gets you back in case something fails.

knutfrode:
Your recipe seemed to work well!
Merging now.

gauteh:
👍

gauteh:
Getting the following warning when running example_generic:

```
WARNING:py.warnings:/Users/gauteh/dev/drift/opendrift/opendrift/readers/basereader.py:380: RuntimeWarning: invalid value encountered in less
  np.array(env[variable]) < np.float(standard_names[variable]['valid_min']),

WARNING:py.warnings:/Users/gauteh/dev/drift/opendrift/opendrift/readers/basereader.py:381: RuntimeWarning: invalid value encountered in greater
  np.array(env[variable]) > np.float(standard_names[variable]['valid_max'])))[0]
```

knutfrode:
I have a fix for the above, using ```numpy.clip()```
Should I make a new PR with this?

knutfrode:
No sorry, clip() is not a fix here...

knutfrode:
But now I have another fix. Making a new PR for this one.

knutfrode:
See https://github.com/OpenDrift/opendrift/pull/251

PR #252, opened 2020-03-02
Try older commit to see docs-build duration

gauteh:
See if this is faster than newer version.


gauteh:
> 3.5 h for approx 20 commits back but with verbosity.

gauteh:
Ser ut til at det er interpolation eksempelet som tar so lang tid at uten noko output so går det over timeouten uten output på 40 minutt, og dermed feiler det. Då er det eit eller anna som har skjedd som gjer at det plutselig går treigare.

gauteh:
example_interpolation takes about 1.5 hours on circleci, seems like most of the time is spent in initializing interpolator. Is the plan to remove the code entirely or should we try to fix it?

knutfrode:
> example_interpolation takes about 1.5 hours on circleci, seems like most of the time is spent in initializing interpolator. Is the plan to remove the code entirely or should we try to fix it?

I had planned for some time to remove the two obsolete interpolators (ndimage and linearND), so that only the default and fastest option remains (linearNDFast).

The only thing would be if we would like to understand the mystery why exactly this part seems to be so much slower on CircleCI.
But I would suggest that we simply deactivate example_interpolation now, and then I can also remove the interpolators soon.

gauteh:
Yes, i agree. Let's at least solve the circleci issue so that it doesn't
hold us back from other tasks.

ons. 4. mar. 2020, 11:03 skrev Knut-Frode Dagestad <notifications@github.com

knutfrode:
Ok, then we could simply delete example_interpolation.
Should I make that, eventually as a separate PR? Or would you like to do it?

gauteh:
No, feel free. Won't have the chance for a little while.

ons. 4. mar. 2020, 13:08 skrev Knut-Frode Dagestad <notifications@github.com

knutfrode:
Deleted directly on master here: https://github.com/OpenDrift/opendrift/commit/3476d3caeae3a6c1097d786d093463265d8009a6

knutfrode:
Then I guess we can close this PR?

gauteh:
Yup!


ons. 4. mar. 2020, 14:04 skrev Knut-Frode Dagestad <notifications@github.com

PR #253, opened 2020-03-03
Radionuclide

magnesim:


knutfrode:
This looks ok to me. Agree to merge, @gauteh ?
Doc-build timed out, but I guess that is independent of this commit.

gauteh:
I fixed the doc builds on master, so they should complete. There are a couple of merge conflicts which I can try to sort out and put up on a new branch for @magnesim to review. Seems like some of the stuff from before recent refactoring on our part (some removed files / examples) snuck in.

A couple of places I noticed there are some extra trailing whitespaces, and some places use of `print()` rather than logging framework.

I noticed you put a couple of files in .gitignore. For non-generic files you can put paths in `.git/info/exclude` if you have folders/files in the repo directory that are only useful to you.

gauteh:
There are several `WARNING: Only NaN input to linearNDFast` when running the example, is that intended?

knutfrode:
> There are several `WARNING: Only NaN input to linearNDFast` when running the example, is that intended?

Yes, these are ok, normally indicating that the ocean model does not have deep enough layers for the given element collection.

knutfrode:
> A couple of places I noticed there are some extra trailing whitespaces, and some places use of `print()` rather than logging framework.
> 
For development and debugging it is ok to use ```print```, but generally it is recommended to use ```self.logger.debug``` (or ```info``` / ```warning```), as these then are collected chronologically in the output log-file. Othwerwise print-output are collected at the end.

gauteh:
I put a rebased version in #254. This gets rid of the merge-commits, and hopefully I didn't revert any other changes. The example radionuclide script runs. I got one error when running tests locally, but seems unrelated.

magnesim:
What are the commands to check out the new pull request?

magnesim:
> What are the commands to check out the new pull request?

Ok, I found it. To me, it seems like version #254 works perfectly fine

gauteh:
Merged via #254.

PR #254, opened 2020-03-03
Radionuclide

gauteh:
Rebased version of #253. @magnesim could you please check if this is correct?


magnesim:
I have tested this version, it seems to work as intended. I cannot see any errors.

gauteh:
Great! It seems to be something in the example script that hangs on CircleCI, but it does not do so locally. I'll try to investigate tomorrow.

PR #259, opened 2020-03-06
Extended plot_oil_budget() to (optionally) also plot oil viscosity an…

knutfrode:
…d density, and wind and waves. Programmed by Johannes Rohrs for Ocean Science paper: https://doi.org/10.5194/os-14-1581-2018. Also added conversion of cftime to datetime, to avoid problems with plotting time axis.


knutfrode:
For some reason, only the animation is shown from this example, and not the plot:
https://opendrift.github.io/gallery/example_oil_budget.html
The same for several other example scripts.

PR #265, opened 2020-03-18
ci-imagemagick

gauteh:


gauteh:
@knutfrode: is it ok to remove the old matplotlib support here now?

knutfrode:
Yes, I believe that is very much outdated now.
But we also make imagemagick an explicit dependency? I believe on Linux it is normally available as default?
https://anaconda.org/conda-forge/imagemagick/

gauteh:
I wasn't aware that conda had imagemagick as well, that is good. ffmpeg had
issues when there are several animations generated. we might have to add it
to our list dependencies?

ons. 18. mar. 2020 kl. 13:50 skrev Knut-Frode Dagestad <
notifications@github.com>:

knutfrode:
It was also new to me.
ffmpeg is already in environment.yml
Se yes, seems like a good idea to add imagemagick.

knutfrode:
Both animation and plot is not available on https://opendrift.github.io/gallery/example_oil_budget.html
Was it imagemagick that solved this issue?

gauteh:
I think so. I think on CI the imagemagick was not available, so fell back
to ffmpeg. ffmpeg could not handle multiple animations simultaneously (or
multiple plots?) since I had to remove some plt.close(..) in order to keep
the oil budget open (so that sphinx would detect it). Now installed
imagemagick through ubuntu (on CI), not through conda. Would be best to
keep imagemagick as optional dep, but I dont think that is possible with
environment files.

tor. 19. mar. 2020 kl. 09:35 skrev Knut-Frode Dagestad <
notifications@github.com>:

PR #270, opened 2020-04-08
Read Vtransform from native ROMS files

simonweppe:
Instead of assuming Vtransform = 1 as per default in sdepth()

Using `np.asarray(self.Dataset['Vtransform'])` seems to work whether the "Dataset" is loaded using netCDF4 or xarray.

### DIFF
--- a/opendrift/readers/reader_ROMS_native.py
+++ b/opendrift/readers/reader_ROMS_native.py
@@ -279,14 +279,14 @@ class Reader(BaseReader):
                     self.Dataset.variables['h'][:]
 
                 Htot = self.sea_floor_depth_below_sea_level
-                self.z_rho_tot = depth.sdepth(Htot, self.hc, self.Cs_r)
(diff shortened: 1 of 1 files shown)


knutfrode:
The commit seems ok, but for some reason CircleCI is failing. Do you understand the error message, @gauteh ?
https://app.circleci.com/pipelines/github/OpenDrift/opendrift/565/workflows/32c2e03f-faa9-4501-bf13-c808326949c0/jobs/1588

Also for some reason Travis has not been run here.

knutfrode:
It was indeed a bug that Vtransform was hardcoded as 1, and in fact even one of the included sample datasets had Vtransform=2! So very good that you spotted this.
I therefore also had to update one test to comply with this. I accidentally merged your request before fixing the failing test, but fixed the test afterwards. So everything should be ok now.

I also followed this procedure so that Travis is now again connected:
https://travis-ci.community/t/travis-ci-build-no-longer-showing-as-check-on-github-pull-request-page/6955/7

PR #283, opened 2020-05-14
Release v1.2.1

gauteh:
Is there anything more we should include before releasing 1.2.1?


knutfrode:
No, this should be fine.

PR #381, opened 2020-09-27
allow selection of sub-domain for LCS calculation

johannesro:
- allow to define a subgrid (domain) for the calculation of LCS
- allow to calculate LCS with particles seeded at specified depth
- introduce switch to deactivate plotting of initial particle positions in o.plot. This is useful to check if LCS calculations are realistic.

At present, calculate_ftle uses the map projection of a specified reader. Code could be expanded to also allow the user to provide a proj4 string og pyproj object.

Note that LCS does not work well for fake projections (reader_ROMS_native)


knutfrode:
When later allowing user to provide proj4-string, then latlong could be default, so that user can provide lon/lat-values, which are more human-readable than x,y. In this case, I believe it should be no problem to use e.g. reader_ROMS_native for the drift calculations.

PR #391, opened 2020-10-06
allow proj object or string to be passed to calculate_ftle

johannesro:
elaborating on calculate_ftle; allow custom projections to be passed by user.


knutfrode (approved):
Seems good.

PR #392, opened 2020-10-07
Allow for constant 'ocean_vertical_diffusivity' values

simonweppe:
Adding a new option for `vertical_mixing:diffusivitymodel` = 'constant'. where the fallback value is used.


knutfrode:
I accepted your pull request, but later realised that if there are any readers added which actually provide diffusivity valyes, then the constant value (fallback_value) does not override these.
Thus I made I new commit now, where constant values overrides any environment-values.
https://github.com/OpenDrift/opendrift/pull/393/files#diff-25aee4a83f4d5b3823736d0e1cd98e43
At the same time I removed diffusivitymodel `zero` which is just a special case of constant.

simonweppe:
Yes good point. Thanks.
________________________________

PR #430, opened 2020-11-13
Radionuclide

magnesim:


knutfrode:
Looks like many example-scripts have disappeared here?
https://github.com/OpenDrift/opendrift/pull/430/files

Ideally, this PR should only affect radionucleides.py and example_radionucleides.py

@gauteh might be able to help with Git-technicalities

magnesim:
> Looks like many example-scripts have disappeared here?
> https://github.com/OpenDrift/opendrift/pull/430/files
> 
> Ideally, this PR should only affect radionucleides.py and example_radionucleides.py
> 
> @gauteh might be able to help with Git-technicalities

Could it be renaming of example scripts a while ago? Or - it could be mistakes by me...

gauteh:
I'd be happy to help. It would be good to do `git rebase` to get rid of all the merge commits as well. I ran `pylint -E opendrift` and encountered the following errors in this module: 

```
************* Module opendrift.models.radionuclides
opendrift/models/radionuclides.py:559:26: E0602: Undefined variable 'interp1d' (undefined-variable)
opendrift/models/radionuclides.py:1284:16: E1101: Instance of 'RadionuclideDrift' has no 'conc_topo' member (no-member)
opendrift/models/radionuclides.py:1289:18: E1101: Instance of 'RadionuclideDrift' has no 'conc_lat' member (no-member)
opendrift/models/radionuclides.py:1290:18: E1101: Instance of 'RadionuclideDrift' has no 'conc_lon' member (no-member)
``` 

I think the first one at least is real. The last ones are probably because the members are not defined at class level, this is better practice (even if just defined as None).

knutfrode:
Yohoo

PR #436, opened 2020-11-17
Unstructured reader (FVCOM)

gauteh:
Rember:
- [ ] Check if wind_from_speed_and_direction works
- [x] Remove `use_block`
- [ ] Check if changes to openberg regarding use_block and return_block causes trouble


knutfrode:
Interesting rearrangement.

gauteh:
Yes, renamed analytical to continuous as that seems to make more sense. Not all models using it are analytical, but they can provide values at continuous input. Unstructured will cover infinite dimensions, but they will mostly be automatically garbage collected by python.

knutfrode:
Continuous readers will mostly be continuous in both time and space, though there may be exceptions. E.g. new reader_constant_2d is continuous in time, but discrete/gridded in space.

gauteh:
Yes, but it is continuous from the point of view of basemodel - it can take any input value and does its own interpolation (or selects nearest value)?

gauteh:
If removing `return_block`, `use_block` should maybe also be removed and left to be decided by the readertype (structured, continuous)?

knutfrode:
On second thought, [reader_constant_2d](https://opendrift.github.io/_modules/opendrift/readers/reader_constant_2d.html#Reader) should be a regular structured reader, as it does not do its own interpolation, but returns a block.

So I guess what characterizes `analytical` or `continuous` readers is that they do their own interpolation directly onto particle positions/times directly at each call?

Another reader in the borderland is [reader_ArtificialOceanEddy](https://opendrift.github.io/_modules/opendrift/readers/reader_ArtificialOceanEddy.html#Reader), which uses an analytical equation to calculate variables directly at particle positions, but which can also return this on a discrete grid if user has specified `use_block = True`
I guess we should here chose one of the options, perhaps the former(?)

gauteh:
> On second thought, [reader_constant_2d](https://opendrift.github.io/_modules/opendrift/readers/reader_constant_2d.html#Reader) should be a regular structured reader, as it does not do its own interpolation, but returns a block.

Yes, exactly.

> So I guess what characterizes `analytical` or `continuous` readers is that they do their own interpolation directly onto particle positions/times directly at each call?

Yes, indeed. So you can ask for values for a continuous range of arguments. I am open for both names, it is important that we describe these things in the documentation of the class.

> Another reader in the borderland is [reader_ArtificialOceanEddy](https://opendrift.github.io/_modules/opendrift/readers/reader_ArtificialOceanEddy.html#Reader), which uses an analytical equation to calculate variables directly at particle positions, but which can also return this on a discrete grid if user has specified `use_block = True`
> I guess we should here chose one of the options, perhaps the former(?)

Is there any advantage in allowing `use_block`? If not, then the choice is simple.

> And then there is the case that reader provides interpolated variables, but with discrete times, such as [reader_timeseries](https://opendrift.github.io/_modules/opendrift/readers/reader_timeseries.html#Reader)
> This one presently does it own interpolation in time (actually simply nearest in time).
> I guess it makes sense to require that readers which do interpolation in space, should also do interpolation in time (whenever they have a discrete time axis).

That is the current constraint. I think the most natural solution, if following the file/class layout in this PR, is to have StructuredSpatial and StructuredTemporal classes so that they can be composed. Have to think about how exactly that can work, one might have to inherit the other. Maybe we have to go for ContinuousTemporalStructuredSpatial.

knutfrode:
I agree that `use_block`, can be removed. It was probably just a bad idea a long time ago.

gauteh:
> I agree that `use_block`, can be removed. It was probably just a bad idea a long time ago.

I don't know about that, but if it is not used anymore it would make things simpler to remove it.

gauteh:
Cannot use xarray (yet) because of: https://github.com/pydata/xarray/issues/2233

gauteh:
See https://github.com/gauteh/opendrift/tree/fvcom-noblock for code removing `block` and the failing tests.

gauteh:
Notes on current implementation:
* Uses nearest neighbor in time and space
* Slicing on thredds causes too many opendap-requests, see if `.diagonal()` is better for remote sources. Will probably always be an issue.
* animations does not work, because basemodel probably need to use `get_variables_interpolated_xy(..)`
* Seems to work for local files.

opendrift/readers/reader_netCDF_CF_unstructured.py
```diff
@@ -246,11 +246,16 @@ def get_variables(self,
         * salinity
 
         """
-        logger.debug("Requested variabels: %s, lengths: %d, %d, %d" %
-                     (requested_variables, len(x), len(y), len(z)))
 
         x = np.atleast_1d(x)
         y = np.atleast_1d(y)
+        z = np.atleast_1d(z)
```
gauteh, line 252: for `x` and `y` this is already done in variables.py:get_variables_interpolated_xy

opendrift/readers/reader_netCDF_CF_unstructured.py
```diff
@@ -284,9 +285,12 @@ def get_variables(self,
                 sigmas = self.__nearest_node_sigma__(dvar, nodes, z)
                 assert len(sigmas) == len(z)
 
-                # for some unfathomable reason netCDF4 uses different fancy-indexing than numpy.
-                indx_t = np.repeat(indx_nearest, len(sigmas))
-                variables[var] = np.array([ dvar[ti, si, ni] for ti, si, ni in zip(indx_t, sigmas, nodes) ])
+                # Reading the smallest block covering the actual data
+                block = dvar[indx_nearest,
+                             slice(sigmas.min(), sigmas.max()+1),
+                             slice(fcs.min(), fcs.max()+1)]
```
gauteh, line 298: `fcs` not defined.

opendrift/readers/reader_netCDF_CF_unstructured.py
knutfrode: Ok, then we can also do the same for z there, and convert it to same length if scalar.

opendrift/readers/reader_netCDF_CF_unstructured.py
knutfrode: oh yes, should be "nodes" I guess

opendrift/readers/reader_netCDF_CF_unstructured.py
```diff
@@ -284,9 +285,12 @@ def get_variables(self,
                 sigmas = self.__nearest_node_sigma__(dvar, nodes, z)
                 assert len(sigmas) == len(z)
 
-                # for some unfathomable reason netCDF4 uses different fancy-indexing than numpy.
-                indx_t = np.repeat(indx_nearest, len(sigmas))
-                variables[var] = np.array([ dvar[ti, si, ni] for ti, si, ni in zip(indx_t, sigmas, nodes) ])
+                # Reading the smallest block covering the actual data
```
gauteh, line 295: Yes, can be moved out. But need to communicate whether face or node. Better to do so explicitly than with checking dimension. Maybe send siglay or siglay_center as argument.

opendrift/readers/reader_netCDF_CF_unstructured.py
gauteh: Yes, reason I left it was in case logic in between required z = None for something. Think some readers choose depth themselves.

gauteh: This will be easier to decouple when we move `get_variables_impl` and `get_variables_derived` either above or below in the reader-stack.

opendrift/readers/reader_netCDF_CF_unstructured.py
gauteh: Hm, not face or node - but siglay or siglevel.

opendrift/readers/basereader/unstructured.py
```diff
@@ -0,0 +1,194 @@
...
+                                      reader_x,
+                                      reader_y,
+                                      z)
+
+        # We probably have to use an UnstructuredBlock to store closest time-steps, and thus avoid fetching
+        # more data on every call.
+
+        logger.debug('Fetched env-before')
+        env_profiles = None
+        if profiles is not None:
```
knutfrode, line 78: This is presumably a temporary fix?

gauteh: It was probably copied from the code that previously handled analytical readers.

gauteh:
Converting the FVCOM file from Netcdf3 to Netcdf4 doubles the speed for local files, and will probably make it easier for opendap servers as well.

knutfrode:
After the two last commits, I expect all examples to complete.

gauteh:
Re-enabled tests, fixed a couple of errors. Removed traces of block I could
find. Timeseries at position values was wrong because old code use block =
False which causes different interpolation. With block = True in old code
values match exactly.

man. 30. nov. 2020 kl. 19:27 skrev Gaute Hope <gauteh@met.no>:

> Great! I am running now locally too to test.
>
>>

gauteh:
Great! I am running now locally too to test.

>

PR #440, opened 2020-11-20
docs: move front page example to example file in order to avoid re-run every time

gauteh:


knutfrode:
Should this be merged, or closed?

gauteh:
Ah, this should be merged. I forgot it.

PR #445, opened 2020-11-26
reader_netCDF_CF_generic is now based on Xarray as only option

knutfrode:


gauteh:
See this: Cannot use xarray (yet) because of: pydata/xarray#2233, prevents the use of xarray with fvcom files. Will not open at all.

knutfrode:
Oh, yes....
But I don't think this is a problem for the netCDF_generic reader?

gauteh:
No, not unless it encounters the same type of coordinate-setup. But I don't
think that happens very often. But if it does, it is difficult to avoid at
the moment.

tor. 26. nov. 2020 kl. 10:50 skrev Knut-Frode Dagestad <
notifications@github.com>:

Issue #243, opened 2020-02-13
Improve unstructured netCDF reader (FVCOM, Telemac, ...)


knutfrode:
The [unstructured netCDF reader](https://github.com/OpenDrift/opendrift/blob/master/opendrift/readers/reader_netCDF_CF_unstructured.py) is far less developed than the counterpart for regular grids ([reader_netCDF_CF_generic](https://github.com/OpenDrift/opendrift/blob/master/opendrift/readers/reader_netCDF_CF_generic.py)).
Things to improve:
- the present version has no vertical interpolation, and returns [only surface values](https://github.com/OpenDrift/opendrift/blob/master/opendrift/readers/reader_netCDF_CF_unstructured.py#L232).
- present version makes an intermediate horizontal interpolation from unstructured grid onto a regular lon-lat grid. While this is efficient when re-used, it is not an ideal solution, as the unstructured grid normally has strongly varying resolution over the domain. A fast interpolation directly from unstructured grid onto particle positions is probably better. However, this interpolation must then be "hidden" inside the reader, or else [Basereader](https://github.com/OpenDrift/opendrift/blob/master/opendrift/readers/basereader.py)/[interpolation](https://github.com/OpenDrift/opendrift/blob/master/opendrift/readers/interpolation.py)/[ReaderBlock](https://github.com/OpenDrift/opendrift/blob/master/opendrift/readers/interpolation.py#L226) must be updated to also support unstructured ```blocks```. This might be a relatively complicated task.
- the [__init__ method](https://github.com/OpenDrift/opendrift/blob/master/opendrift/readers/reader_netCDF_CF_unstructured.py#L34) contains a lot of duplication from [reader_netCDF_CF_generic](https://github.com/OpenDrift/opendrift/blob/master/opendrift/readers/reader_netCDF_CF_generic.py). Most of this can probably be shared (factored out).

@gauteh might be able to look into this either this spring, or early 2021.

It would then be useful to have available some (small) sample datasets for development.
- Perhaps @johannesro and @simonweppe can provide two sample FVCOM datasets for this purpose?
- Can someone provide a small sample output file from Telemac-model (As far as I know, this also follows the same CF-convention?)
- I believe this is the specifications to follow: http://ugrid-conventions.github.io/ugrid-conventions/

simonweppe:
Hi Knut-Frode, and all , 

[...]

PR #447, opened 2020-11-26
Reader native schism

simonweppe:
Hi @knutfrode @gauteh ,

Following up on that discussion on readers for data on unstructured grids, https://github.com/OpenDrift/opendrift/issues/243, I have made a clean branch from your latest code (without any of my other edits/dev) and added only work I did on the reader for the native SCHISM output files ( SCHISM model : [http://ccrm.vims.edu/schismweb/](). The interpolation is handled by the `ReaderBlockUnstruct()` class, which is in the reader itself rather than in `interpolation.py` for now

Like you said, it would be good to have a more generic reader for several unstructured grid formats, but in the meantime this could be useful to SCHISM users or in the development of a more unified reader for these kind of grids.

Cheers,

Simon


gauteh:
Hi @simonweppe,

Thank you very much. We are in the process of restructuring the readers
module to better accommodate unstructured readers. Since these are more
heterogeneous than regularily gridded, it will be a bit different from the
structured ones. The pr is here:
https://github.com/OpenDrift/opendrift/pull/436 . I think we should merge
the work in this pr afterwards, it should not be too difficult to adapt it.

Regards, Gaute

tor. 26. nov. 2020, 16:16 skrev simonweppe <notifications@github.com>:

gauteh:
Hi @simonweppe,

I've started looking at this reader. I have posted some cleanups in #476 (currently only regarding the custom landmask). I have also rebased your work. We have made some significant changes to the structure and inheritance of readers which we need to take into account in your reader before we can merge. It would be good to remove debug comments and dead code before merging since this tends to end up never getting removed later. If you do not have time to do this I can also try to help out with that. I have not really started working on the schims-reader yet but I hope that some of the methods in unstrucutred.py can be used to avoid duplication between FVCOM and SCHISM. I have some doubts though since you need a 3D cKDTree. 

In custom landmask reader, do you need the plot method? I assume it is duplicated here since you need an unstructured boundary?

Regards, Gaute

simonweppe:
Thanks for that @gauteh. I'll find some time to clean the code and include as much as possible the methods in unstrucutred.py. 
Keep you in touch.
Thanks,
Simon

PR #451, opened 2020-12-01
Dev

knutfrode:


opendrift/timer.py
```diff
@@ -6,9 +6,15 @@ class Timeable:
     Utility class for measuring total time spent in various steps in a class
     throughout program execution.
     """
+
     timers = OrderedDict()
     timing = OrderedDict()
```
gauteh, line 11: These can now be None.

opendrift/timer.py
```diff
@@ -6,9 +6,15 @@ class Timeable:
...
 
+    def __init__(self):
+        # Resetting timers for each object
+        self.timers = OrderedDict()
+        self.timing = OrderedDict()
+
```
gauteh, line 17: We also have to make sure sub-classes call `super().__init__()`.

opendrift/timer.py
knutfrode: Getting a little tricky.
If these are None, `super().__init__()` must be added to both `Basemodel.__init__` and `Variables.__init__`
This is fine, but then there is problem with `netCDF_CF_unstructured`, which needs a timer before `super().__init__()` is called. But this cannot be called until after some data are collected....

gauteh:
The original values of those variables will be overwritten anyway, so that
needs to be solved some other way. Maybe it can be done with cached
properties.

tir. 1. des. 2020, 17:03 skrev Knut-Frode Dagestad <notifications@github.com

opendrift/timer.py
gauteh: True, see https://github.com/OpenDrift/opendrift/pull/453 for a suggestion on how to do it with properties and avoiding using an __init__. Even with __init__ the original dict would be overwritten if used before __init__'ed.

knutfrode: Yes, this is more elegant.

knutfrode: Then we should remove this commit from this PR, but keep the other one (subset of times when importing).
Can that be done through GitHub?

gauteh:
On your dev-branch you do:

$ git rebase -i master
delete the commit-line with your timer-stuff
$ git push -f


ons. 2. des. 2020 kl. 08:15 skrev Knut-Frode Dagestad <
notifications@github.com>:

PR #472, opened 2020-12-14
Removed obsolete lines from reader_ROMS_native

knutfrode:


knutfrode:
@gauteh I believe there are more such things which can be removed from various readers, or moved up to Basereader.
Several readers prepare xmin, delta_x etc, but I believe the task of readers can be simplified to returning either:

- arrays of `x` and `y` plus a `proj4` string (x/y will be lon/lat if proj is latlong)

 or

- arrays of `lon` and `lat`. If these are 1D, proj shall be set to `proj=latlong` in Basereader, and x=lon, y=lat. If these are 2D, projected shall be set to False, and 2D interpolators `xy<-> lonlat` shall be made.

gauteh:
Yes, I agree. I was thinking about the same thing. Either projected or unprojected grids should be generated in the same way. This functionality may belong in ReaderDomain, or it may belong in structured if it is only relevant to structured grids. Also slicing and indexing can be generalized, as I have started in readre_grib2. The most general one is in netcdf-generic so the other readers should also benefit from e.g. date-line crossing and correct slicing.

knutfrode:
Do you understand why docs-built fail on this commit? It works well on your other PR.

gauteh:
Yes, I fixed it in my PR. Just commited the specific fix to master, so should work if you re-trigger on this PR.

knutfrode:
I believe readers should also not have to define `start_time` and `time_step` etc, as these could be `None` by default
This would simplify e.g. this constant_2d-reader (Contineous):
https://github.com/OpenDrift/opendrift/blob/master/opendrift/readers/reader_constant_2d.py

And this one could perhaps also store `self.x` instead of `self.xmin` and `self.xmax`, which can be calculated in parent class.

PR #473, opened 2020-12-14
Unwrap get_variables_wrapper (impl) + get_variables_derived so that these are not called back from sub-class

gauteh:
- reader: unwrap get_variables_impl and get_variables_derive
- pytest: register slow and veryslow marks

### DIFF
--- a/opendrift/readers/basereader/continuous.py
+++ b/opendrift/readers/basereader/continuous.py
@@ -1,3 +1,4 @@
+from abc import abstractmethod
 import numpy as np
 from .variables import Variables
 from .consts import vector_pairs_xy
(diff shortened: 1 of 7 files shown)


gauteh:
I don't understand why `test_valid_minmax_nanvalues` does not work, or how exactly it is supposed to work.

knutfrode:
This test tries to assure that read environment variables larger than min/max are replaced with the corresponding min/max value.
So this test sets max x-current to 0.1 m/s, and checks that the x-current returned has no larger values than this.

but unclear to me why it fails after this commit.

gauteh:
> This test tries to assure that read environment variables larger than min/max are replaced with the corresponding min/max value.
> So this test sets max x-current to 0.1 m/s, and checks that the x-current returned has no larger values than this.
> 
> but unclear to me why it fails after this commit.

It checks the exact value, not that it is less than. But it still should not change. I see that the values are set to 0, is that incorrect? Weird that the test above works.

knutfrode:
Yes, values larger than the limit (here 0,1) should be set to exactly 0.1. And this tests checks that the max values of x-current is exactly 0.1. But quite strange that it suddenly is 0 in this case.

gauteh:
Ok, i'll keep looking

gauteh:
Could it be because the check now is done after reader block interpolation?

man. 14. des. 2020, 14:08 skrev Knut-Frode Dagestad <
notifications@github.com>:

knutfrode:
But `o.get_property()` should have provided data after interpolation also before this commit?

Anyway it should probably not return exactly 0?

PR #474, opened 2020-12-14
Removed obsolete lines in reader_ROMS_native

knutfrode:


knutfrode:
The only failure was due to unavailable thredds server, so merging.

PR #477, opened 2020-12-15
Fixed reader.plot() by replacing __get_variables_derived__ with get_v…

knutfrode:
…ariables, and added unit test. Fixed misprint (convole). Removed some obsolete(?) checking in reader_netCDF_CF_generic

### DIFF
--- a/opendrift/readers/basereader/__init__.py
+++ b/opendrift/readers/basereader/__init__.py
@@ -309,14 +309,13 @@ class BaseReader(Variables):
         if variable is not None:
             rx = np.array([self.xmin, self.xmax])
             ry = np.array([self.ymin, self.ymax])
-            data = self.__get_variables_derived__(variable, self.start_time,
-                                      rx, ry)
+            data = self.get_variables(variable, self.start_time, rx, ry)
             rx, ry = np.meshgrid(data['x'], data['y'])
             rx = np.float32(rx)
             ry = np.float32(ry)
             rlon, rlat = self.xy2lonlat(rx, ry)
             data[variable] = np.ma.masked_invalid(data[variable])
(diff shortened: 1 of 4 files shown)


opendrift/readers/basereader/__init__.py
```diff
@@ -309,14 +309,13 @@ def plot(self, variable=None, vmin=None, vmax=None,
         if variable is not None:
             rx = np.array([self.xmin, self.xmax])
             ry = np.array([self.ymin, self.ymax])
-            data = self.__get_variables_derived__(variable, self.start_time,
-                                      rx, ry)
+            data = self.get_variables(variable, self.start_time, rx, ry)
```
gauteh, line 312: This method is not guaranteed to be available. It will also give different result depending whether your're using a structured or unstructured reader (same as before). The only guaranteed methods, which is also always work in the same way, are `get_variables_interpolated` and `get_variables_interpolated_xy`.

knutfrode: Ok, but then also the reader.plot() method needs to be rebuilt. So this would be a quick temporal fix.

gauteh: Sounds good 👍

knutfrode: It is desirable here to show the reader background data in native resolution ("rotated raw pixels"). With interpolated data, some tricks will be needed to achieve this.

knutfrode: And actually that means using nearest interpolation, which I believe is not available(?) Thus perhaps we can use some method based on top of `get_variables` But adapted to the three types of readers.

PR #479, opened 2020-12-22
Update to LCS algorithms

johannesro:
- bugfixes
- return eigenvector and eigenvalue to calculate strainlines
- distinguish between LCS and FLTE
- add example script for strainline calculation


opendrift/models/basemodel.py
```diff
@@ -1753,11 +1753,11 @@ def seed_within_polygon(self, lons, lats, number=None, **kwargs):
         area = 0.0
         for i in range(-1, len(x)-1):
             area += x[i] * (y[i+1] - y[i-1])
-        area = abs(area) / 2
 
         # Make points, evenly distributed
         deltax = np.sqrt(area/number)
         lonpoints = np.array([])
+        area = abs(area) / 2
```
knutfrode, line 1760: Seems generally good, but 4 tests are failing, presumably because this line has been moved, and thus affecting the area and giving the wrong number of elements. Is this a mistake?

opendrift/models/basemodel.py
```diff
@@ -2043,7 +2043,7 @@ def set_fallback_values(self, refresh=False):
 
     def run(self, time_step=None, steps=None, time_step_output=None,
             duration=None, end_time=None, outfile=None, export_variables=None,
-            export_buffer_length=100, stop_on_error=False):
+            export_buffer_length=100, stop_on_error=False, move_from_land=True):
```
knutfrode, line 2046: `move_from_land` seems to be a duplicate of existing config option `seed:ocean_only`? Thus I believe `move_from_land` should be removed, and config should be used instead, for consistency.

johannesro: I don't remember exactly why I did this, but it was necessary to get it work, and it is different from seeding

opendrift/models/basemodel.py
```diff
@@ -2357,15 +2357,17 @@ def run(self, time_step=None, steps=None, time_step_output=None,
...
-            self.timer_start('preparing main loop:moving elements to ocean')
-            self.elements_scheduled.lon, self.elements_scheduled.lat = \
+
+        # Move point seed on land to ocean
+        if move_from_land==True:
+            if self.get_config('seed:ocean_only') is True and \
+                ('land_binary_mask' not in self.fallback_values) and \
+                ('land_binary_mask' in self.required_variables):
+                self.timer_start('preparing main loop:moving elements to ocean')
+                self.elements_scheduled.lon, self.elements_scheduled.lat = \
```
knutfrode, line 2367: It seems here that you have simply overwritten the config `seed:ocean_only` with the new `move_from_land=True`, and this is probably one (of several?) reason for the tests failing.
Anyway, if starting from a fresh, rebased fork, it should be quite straightforward to make a new PR with the same additions as in this PR. So two things are essential: rebasing, and running pytest before committing.

opendrift/models/basemodel.py
knutfrode: We should avoid more input parameters to methods like `run`, and instead control such things via the config-mechanism. There are several reasons for this, one being that web interfaces like Drifty are based generally on the config-mechanism, and there we want to avoid any method-specific tweaks that mus tbe hardcoded and kept in sync.

PR #491, opened 2021-01-08
test: (regression) seed_elements backward compat does not work for cone

gauteh:


knutfrode:
Discarding this one, since seed_elements() with keyword cone=True is not anymore allowed after #492.

PR #515, opened 2021-01-21
tests: test ascii format is unchanged

gauteh:
Target file is now the new one. It should be update to match the one before latest changes.


knutfrode:
Ja, dette er nok lurt!

knutfrode:
Men feila også pga NOAA-unavailability.

PR #553, opened 2021-03-04
Checking if pykdtree can speed up map plotting with Cartopy.

knutfrode:


knutfrode:
Issues were only the usual DAP-errors.

Do you think we should merge this issue? Does not seem to make much improvement of runtime, but probably also no harm.

gauteh:
Maybe it would be better to add a section about performance on the documentation page about it. There might be other tips that should go there as well. Then we avoid an extra dependency.

knutfrode:
I agree - number of dependencies should be kept to a minimum.

Then I guess we simply close this issue?

gauteh:
Yes!

man. 8. mar. 2021 kl. 09:22 skrev Knut-Frode Dagestad <
notifications@github.com>:

PR #565, opened 2021-03-23
Seeding from STEAM

manuelaghito:


knutfrode:
This seems all fine, except that I believe the seed-method should belong in chemicaldrift.py and not in basemodel.py, as it is not of interest/use to other models.

PR #583, opened 2021-04-12
animation_profile() with colors and legend

manuelaghito:


knutfrode:
Some tests/examples failed because parts of a thredds-catalog was (and is still) down. We can restart and merge this PR after it is back up, hopefully within a few hours.

PR #599, opened 2021-04-21
add shyfem reader

gauteh:


knutfrode:
Ready to merge?

gauteh:
No, not quite yet.

man. 26. apr. 2021, 08:40 skrev Knut-Frode Dagestad <
***@***.***>:

gauteh:
I think this should be ready now, @Manuel Aghito ***@***.***> do you
think the shyfem-reader is functional and not incorrect as it is?

man. 26. apr. 2021 kl. 09:01 skrev Gaute Hope ***@***.***>:

gauteh:
@manuelaghito : can you check if the latest edition makes sense. This uses the middle of layer rather than the bottom.

manuelaghito:
> @manuelaghito : can you check if the latest edition makes sense. This uses the middle of layer rather than the bottom.

I guess you are right, since level description is "bottom of vertical layers" this should correct, I think we can merge this

PR #613, opened 2021-04-30
Dev

manuelaghito:
Fixing shyfem test and example scripts for new sample data


manuelaghito:
Ok, this is replaced by a new PR without double commit

PR #628, opened 2021-05-26
telemac

gauteh:
- Moved the variable conversion at the script initialisation added an example to read a selafin file added a selafin file output from the particle example of Telemac 3D
- telemac: add tests file
- test_readers: convert to pytest
- variables: add center method to find center of reader domain
- telemac: clean up a bit
- circleci: disable tests requiring telemac


Boorhin:
Thanks for that

PR #634, opened 2021-05-30
[run-ex] Illustrating use of orign_marker in example_multi_seed.py

knutfrode:


gauteh:
By the way, it should not be necessary to specify [run-ex] in this case
since you are changing an examples script.

søn. 30. mai 2021 kl. 17:28 skrev Knut-Frode Dagestad <
***@***.***>:

PR #640, opened 2021-06-11
landmask: test using roaring-landmask if available

gauteh:


gauteh:
@knutfrode any objections to including this, as long as roaring is not installed it will use opendrift-landmask-data as usual. We can take it back out again if roaring-landmask is abandoned.

knutfrode:
Very interesting, no objections. What are the benefits? Less memory/disk consumption? How about performance?

gauteh:
Yes, much less memory and disk-consumption goes down by by 3.7GB to about 60mb. Performance is slightly worse, but probably insignificant: 140ms @ 250k points vs 120ms @ 250k points.

knutfrode:
Peis pao.

PR #647, opened 2021-06-25
Updated version of SCHISM reader

simonweppe:
Hi there,

Just sharing some progress on the reader for native SCHISM netcdf files (on unstructured mesh). 

This version aims to be more consistent with the recent Opendrift releases, re-uses some of the methods in `unstructured.py` and tries to better follow the new logics. 

Note the interpolation approach is similar to previous version and close to what is done in `structured.py` with cached reader "blocks" that are re-used for successive interpolation between two model time steps (rather than checking out data at each new interpolation step).

Not in state ready for direct ingestion but keen to have your initial feedbacks/directions to adjust things. 

(Note SCHISM can be coupled with a wave model WWM-II on the same unstructured mesh. I've had a go at a reader for these output netcdf files as well : here [https://github.com/simonweppe/opendrift/blob/master/opendrift/readers/reader_netCDF_CF_unstructured_WWM.py](url) )


knutfrode:
Good work, getting closer!
At least technically, it is now possible to merge this.
However, there is a large amount of duplication, which I commented a few places (inline comments above).
Some of this should probably be avoided (e.g. overloading/duplicating the plot methods), and some of it (e.g. overloading the block-interpolation) should be moved up one step to the unstructured module (though this would need some adaptation in all unstructured readers).
It is still possible to merge this now/soon, but then there should be a goal to gradually reduce the duplication, as the code will otherwise be very difficult to maintain and develop further: https://wiki.c2.com/?DontRepeatYourself

Another small concern is the shoreline example, which is about  0.7MB. If this is added to the repository, it will grow by this size forever, as it remains in history, even if removed in the future. Thus, it would be good to reduce this one to a minimum size in order to illustrate the point/reader.

Note that there is in the meantime now 3 other unstructured readers:
- https://github.com/OpenDrift/opendrift/blob/master/opendrift/readers/reader_telemac_selafin.py
- https://github.com/OpenDrift/opendrift/blob/master/opendrift/readers/unstructured/shyfem.py
- https://github.com/OpenDrift/opendrift/blob/master/opendrift/readers/reader_netCDF_CF_unstructured.py  (adapted to FVCOM)

simonweppe:
Ok thanks @knutfrode  - yes I agree, quite a few things to tidy up before it is in a merging state . I was keen for your initial feedback and directions. I will refine when I find a window and keep you posted.

Great that the reader's collection is growing, I will have a look as I !

simonweppe:
Hi @knutfrode , just a quick update that I've trimmed the pull request to focus on the reader_schism_native. This should make our exchanges easier (i.e. there are no more reference to the landmask custom, I'll make another pull request for this one). I have shortened the reader name too. Note The `ReaderBlockUnstruct `class is still in `reader_schism_native.py` as it's only used for that reader for now but happy to move it if preferred.

gauteh:
Maybe we should move all these readers into the `unstructured` folder?

simonweppe:
@knutfrode , @gauteh 

It seems that I need to add a `covers_positions_xy()` function in the reader if I want to overload only `_get_variables_interpolated_` and get the code to then use the `get_variables_interpolated_xy` function from `variables.py` (which then calls `_get_variables_interpolated_()` internallly). 

Or am I missing something ?

knutfrode:
I am not sure if I understand the last question.

Anyway, the commit now only contains the SCHISM reader, and a dedicated example script, so that it can be merged without danger of breaking other things.
Thus, Gaute and I can merge this now, if ok with you?

There is still a very large amount of duplication/overloading, as this reader is more than 1100 lines, whereas most other readers are a couple of hundred lines. Much of this is however due to the ReaderBlockUnstruct class, which it would be nice to eventually push "upwards" to the unstructured class. But probably we can just merge this now, and move things "upwards" later whenever we find time for it.

knutfrode:
Btw, the two failing tests are due to a temporary unavailability of an http-server with shapefiles, which can be neglected. Otherwise all tests seem to pass now.

simonweppe:
@knutfrode - yes ok with me. Im still working on some features on the side but it can be later pull requests. Indeed I realize it is quite a big piece of code compared to other readers (because of `ReaderBlockUnstruct` as you noted..I wanted to keep all new codes separate from your codebase to make it easy to review, but of course happy for you to rearrange how you see fit).

About my question above, it applies to unstructured readers only.
From what I understood, best practice when creating a new reader would be to include a `_get_variables_interpolated_` function, as suggested [here](https://github.com/OpenDrift/opendrift/blob/master/opendrift/readers/basereader/variables.py#L580). In variables.py, this function gets called inside `get_variables_interpolated_xy` [here](https://github.com/OpenDrift/opendrift/blob/master/opendrift/readers/basereader/variables.py#L667). `get_variables_interpolated_xy` also calls `covers_positions_xy` which, unless is overloaded by the new reader, will use the version in `variables.py`. This would be ok for structured readers, but in the case of unstructured readers, that would not make use of the `self.boundary` for in-mesh checks (which captures the semi-circular outer boundary of model domain). 

But confusion might come from my side, and was just keen to raise the question..I came across the issue when I realized my out-of-domain particles were not being flagged properly (i.e. not capturing the semi-circular outer boundary of unstructured reader, but rather just the regular box bounding it).

knutfrode:
Yes, I see now the comment (from @gauteh) that `_get_variables_interpolated_ _must_ be implemented by every reader`
At least for structured readers, this is not the case, they normally only implements the basic `get_variables`.

I do not have a complete overview of the unstructured logics (code structure), which was implemented by Gaute, and perhaps in that case it is true that `_get_variables_interpolated_ must be implemented by every reader`?

gauteh:
Yes, for structured readers the scaffolding is more general so there the `_get_variables_interpolated_` is implemented in a common way, but requiring each _structured_ reader to implement `get_variables`. For unstructured readers it was not so simple to make this general, but the work done here on UnstructuredReaderBlock can be a part of that in the future.

gauteh:
I agree that we should merge this now, as long as it does not affect existing code. That will make it easier to work with.

knutfrode:
Ok, this makes sense, and is what I vaguely believed. Then merging now.

knutfrode:
I ran `example_schism_native.py` locally, and it worked fine.

PR #691, opened 2021-09-02
Chemicaldrift updates for degradation and volatilization

manuelaghito:


knutfrode:
In principle, basemodel should be chemically(!) free of any module specific things, such as "mass_volatilized" here. But Im fine with being a little pragmatic here.

knutfrode:
For me its ok to keep it as it is. A "proper" solution would involve more hassle than is necessary.

gauteh:
Is it all the properties mass, mass_degraded, mass_volatilized? Seems like there already is mass and mass_biodegraded before this PR?

Agree on pragmatic: maybe there should be a check that these parameters exist? Otherwise one way to do this is to create a helper method named e.g.: `plot_get_markersize()` in basemodel and override it in chemicaldrift. Then you use the type-system to avoid any errors. Might be cleaner than a bunch of if'elses. If you don't have time for that now, that's ok.

Issue #690, opened 2021-09-01
Cannot read variable 'sea_floor_depth_below_sea_level' from unstructured FVCOM nc file


qianmosugar:
Dear Developers,

I am working on a FVCOM model for water exchange, which the interaction with bathymetry is directly affect the particle's z position. 
However, I found that **_reader_netCDF_CF_unstructured_** did not actually read the bathymetry data, which provided by input nc file.
Just take the example_fvcom.py file as an example. I can run this code without any problem, however, when I read the output nc file, the variable 'sea_floor_depth_below_sea_level' of all the particles was assigned the default value of 10000. 

![image](https://user-images.githubusercontent.com/86223682/131667093-768e1ef7-18cc-4880-a079-3cc56ef2cc59.png)

Then, I print the model and find that _**Readers not added for the following variables: sea_floor_depth_below_sea_level.**_ However, I print the reader and find that _**sea_floor_depth_below_geoid**_ is in variables.

![image](https://user-images.githubusercontent.com/86223682/131689737-15b06729-53cc-4ef7-9c46-05e11aba536c.png)

If I understand correct, these two variables need a mapping from one to another. I follow the suggestion from Gaute, add the variable alias manually on line 58 (variable_aliases) in reader_netCDF_CF_unstructured.py. Unfortunately, as long as I add the variable alias, the particles do not move. This is because the variables group will error by **_tuple index out of range_** so that x_sea_water_velocity and y_sea_water_velocity are missing. 

Please forgive me for posting this question again. I think share it would help more people.

Looking forward to your reply.
Thanks in advance.
Yachen

knutfrode:
Hi Yachen,

Can you post the full log output from this run, and also an ncdump (the detailed contents) of your netCDF-file?

qianmosugar:
Hi,

The log file of running model and the output of print(reader) and print(model) is attached. 

I use the origin akvaplan file which used in your **_example_fvcom.py_**. The output of ncdump will be the same. 

Thanks in advanced.

[print_model_issue#690.txt](https://github.com/OpenDrift/opendrift/files/7095557/print_model_issue.690.txt)
[run_script_issue#690.txt](https://github.com/OpenDrift/opendrift/files/7095558/run_script_issue.690.txt)
[print_reader_issue#690.txt](https://github.com/OpenDrift/opendrift/files/7095559/print_reader_issue.690.txt)

gauteh:
[...]

PR #692, opened 2021-09-02
reader_netCDF_CF_unstructured can now also read variables with no sig…

knutfrode:
…ma/depth dimension, such as e.g. sea_floor_depth_below_sea_level


knutfrode:
This seems to be the reason for why 2D variables such as ocean depth was not read in e.g. #690
@gauteh please check (tomorrow) if this seems ok

gauteh:
Looks good (it will break for time-dependent vars, but we can handle that later).

PR #696, opened 2021-09-12
Importing reader_global_landmask at top of Basemodel. NB: suspicious failures arising in test_fvcom.py after this

knutfrode:


knutfrode:
Still failing due to the Cartopy http-error. Perhaps deleting CircleCI cache could help by proceeding to updated/fixed version of Cartopy?

gauteh:
I don't see any failures in test_fvcom?

knutfrode:
No, seems like `test_fvcom` is not run on CircleCI ?
But it fails locally, and the last test is also commented as "This is still failing"
https://github.com/OpenDrift/opendrift/blob/master/tests/readers/test_fvcom.py#L130
But strangely, this did not start failing for me until now.

gauteh:
See #697 for skipping those tests when NE shapes are missing. Everything works without the changes in this PR already, was there another reason for moving the imports?

opendrift/models/basemodel.py
```diff
@@ -2786,7 +2783,7 @@ def set_up_map(self, corners=None, buffer=.1, delta_lat=None,
                 logger.debug('Using custom shapes for plotting land..')
                 ax.add_geometries(self.readers['shape'].polys, ccrs.PlateCarree(), facecolor=land_color, edgecolor='black')
             else:
-                opendrift.readers.reader_global_landmask.plot_land(ax, lonmin, latmin, lonmax, latmax, fast, ocean_color, land_color, lscale)
+                reader_global_landmask.plot_land(ax, lonmin, latmin, lonmax, latmax, fast, ocean_color, land_color, lscale)
```
knutfrode, line 2786: For unknown (to me) reason, the above change had to be made.
Then instead of having separate imports of reader_global_landmask throughout Basemodel, I made one at the top. But a but shaky if this change should lead to apparently unrelated failures?

gauteh:
test_fvcom should be run on circleci as well. I don't see the still failing comment?

opendrift/models/basemodel.py
gauteh: Ok, either way should work.

gauteh:
I think you have an outdated master. The circleci config doesn't match the one on master. You could try to rebase this on latest master and force push (_this branch_).

PR #701, opened 2021-09-17
reader_netCDF_CF_generic: bugfix for lazy loading with coordinates in decreasing order

ej81:
The part of the grid corresponding to the active particles is calculated using xmin and ymin, but this should be x[0] and y[0]. For coordinates in decreasing order the minimum is not the same as the first value and the wrong part of the grid would be loaded. The use of np.abs here avoids having to use a signed delta_x and delta_y, which could break code in other places.

### DIFF
--- a/opendrift/readers/reader_netCDF_CF_generic.py
+++ b/opendrift/readers/reader_netCDF_CF_generic.py
@@ -367,8 +367,8 @@ class Reader(BaseReader, StructuredReader):
                 x = np.mod(x, 360)  # Shift x/lons to 0-360 
             elif self.lon_range() == '-180to180':
                 x = np.mod(x + 180, 360) - 180 # Shift x/lons to -180-180
-        indx = np.floor(np.abs(x-self.x[0])/self.delta_x).astype(int) + clipped
-        indy = np.floor(np.abs(y-self.y[0])/self.delta_y).astype(int) + clipped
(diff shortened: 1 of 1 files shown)


knutfrode:
This seems fine, except that it breaks the functionality of clipping boundary pixels from a reader domain:
https://opendrift.github.io/_modules/opendrift/readers/basereader.html#BaseReader.clip_boundary_pixels

With some trial and error, it seems like this can be fixed by replacing your modifications with the following:
```
indx = np.floor(np.abs(x-self.x[0]-self.clipped*self.delta_x)/self.delta_x).astype(int) + clipped
indy = np.floor(np.abs(y-self.y[0]-self.clipped*self.delta_y)/self.delta_y).astype(int) + clipped
```

Can you please commit this modification to your PR, and check that this also solves your initial issue?
If then all tests are also passing, we can merge it.

ej81:
Thanks, I modified your suggestion slightly because otherwise the combination of `clip_boundary_pixels` and a decreasing coordinate variable would not work.

This time I checked also the other tests (not just the ones from `tests/readers/`, my bad...). There is just one failure in `test_leeway.py` when comparing the ASCII output to the reference, but this seems a machine precision issue that is unrelated to this pull request.

knutfrode:
Good, your modification is cleaner, and now all tests are passing. Thus merging now. Thx for the contribution.

The Leeway-precision issue seems to be only on your machine. If this remains a problem, the test can probably be updated to compare with lower precision.

PR #730, opened 2021-10-14
Jul sealice

Boorhin:
An update of the selafin reader with major improvements
Adding the sea-lice module simulating sea-lice behaviour according to the literature with most known parameters affecting particle tracking of super-individuals of sea-lice after hatching


knutfrode:
A new SeaLice model, that is quite interesting! Might be of interest to several users.
With time, it would be nice with an example script demonstrating this one, ideally using the sample datasets within the repository.

Boorhin:
> Good. I think the only remaining issue (as Gaute mentioned in Slack) is that you have modified an old version of the Selafin-reader, whereas Gaute has also modified this reader (in the master branch) in the meantime. So this is why tests are now failing. I don't know whether your modifications are necessary, or already covered by Gautes work? If they are necessary, some merging (further commits to Selafin-reader in this PR) is needed to make all tests pass.

OK I have missed that. The modifications were relative to a major bug I found in the reader so I would need to analyse what he has done

gauteh:
Both split PRs merged.

PR #733, opened 2021-10-15
remove cartopy<0.20 requirement

gauteh:


gauteh:
@knutfrode : can you remember why we had this in the first place?

knutfrode:
No, it seems like you added this 24 days ago, in relation to #708:

https://github.com/OpenDrift/opendrift/commit/ee4db8ef61ea88b0200249e230305f117c7755f6#diff-9efd195f4e9bfb79ccd456a1d8370fafcc4bcb0b00ea3799222667d2ae818533

knutfrode:
But we do get this now:

`pyproj.exceptions.ProjError: Error creating Transformer from CRS.: (Internal Proj Error: proj_create_operations: Source and target ellipsoid do not belong to the same celestial body)`

gauteh:
hm yes, seems like tests fail and we need to make some changes in plotting to work with new cartopy.

knutfrode:
Maybe we can get away with fixes like this:
```
x, max_y = self.transform_point(lon, lat, PlateCarree())
x, max_y = self.transform_point(lon, lat, PlateCarree(globe=self.globe))
```

https://github.com/SciTools/cartopy/pull/1252

PR #739, opened 2021-10-18
Add opendrift command line tool for doing a subset of simulations

gauteh:
- opendrift.py: add opendrift command line
- add opendrift.py to setup.py

@vegardb: Let's see if this happens.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,6 +32,7 @@ setuptools.setup(
                'opendrift/scripts/opendrift_plot.py',
                'opendrift/scripts/opendrift_animate.py',
                'opendrift/scripts/opendrift_gui.py',
-               'opendrift/scripts/mp4_to_gif.bash']
+               'opendrift/scripts/mp4_to_gif.bash',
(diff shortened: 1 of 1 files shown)


opendrift/scripts/opendrift.py
```diff
@@ -0,0 +1,184 @@
...
+        pass
+
+models = {
+    'leeway': get_leeway_model,
+    'openoil': get_openoil_model,
+    'shipdrift': get_shipdrift_model
+}
+
+
+plots = {
```
knutfrode, line 55: This is fine for now, but further down the road it would be nice to generalize this functionality, so that "any" postprocessing method can be applied. That requires some careful thinking.
Then also post-processing methods could be made available through Drifty.

knutfrode:
I had earlier in mind that any OpenDrift module could be called the same way as this script, i.e.
```
>>> openoil.py <options>
>>> leeway.py <options>
```

I don't know whether that is feasible, or whether it is better than having a dedicated script `opendrift.py` as here.

PR #764, opened 2021-11-09
Adding function to check sea surface elevation

simonweppe:
To then be used in surface_stick()

### DIFF
--- a/opendrift/models/physics_methods.py
+++ b/opendrift/models/physics_methods.py
@@ -759,7 +759,27 @@ class PhysicsMethods:
             sea_floor_depth = \
                 env['sea_floor_depth_below_sea_level'].astype('float32')
         return sea_floor_depth
-
+    
+    def sea_surface_height(self):
+        '''fetches sea surface height for presently active elements
+           sea_surface_height > 0 above mean sea level
+           sea_surface_height < 0 below mean sea level
+        '''
+        if hasattr(self, 'environment') and \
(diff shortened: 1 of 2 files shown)


opendrift/models/physics_methods.py
```diff
@@ -759,7 +759,27 @@ def sea_floor_depth(self):
...
+        '''
+        if hasattr(self, 'environment') and \
+                hasattr(self.environment, 'sea_surface_height'):
+            if len(self.environment.sea_surface_height) == \
+                    self.num_elements_active():
+                sea_surface_height = \
+                    self.environment.sea_surface_height
+        if 'sea_surface_height' not in locals():
+            env, env_profiles, missing = \
+                self.get_environment(['sea_surface_height'],
```
knutfrode, line 776: This does not work, since not all sublclasses of OceanDrift (nor OceanDrift itself) have a required variable `sea_surface_height`
Instead this method could return `self.environment.sea_surface_height` whenever it is available, otherwise 0

simonweppe: True ...or maybe it makes sense to have `sea_surface_height` as a required variable for OceanDrift, with fallback zero ? like for wind, and waves.
Happy either way let me know and I can update the pull request accordingly.
Thanks

knutfrode:
I believe this PR is now outdated.
There are some conflicts with recent updates, to probably better to make a new PR if needed.

PR #767, opened 2021-11-16
Improving speed and bugfix in animate_vertical_distribution() 

manuelaghito:
should be ready to merge

Tested on
	example_oil_verticalmixing.py
	example_codegg.py
	example_vertical_mixing.py
	example_oilspill_seafloor.py

### DIFF
--- a/opendrift/models/oceandrift.py
+++ b/opendrift/models/oceandrift.py
@@ -435,7 +435,7 @@ class OceanDrift(OpenDriftSimulation):
         else:
             return None
 
-    def animate_vertical_distribution(self, depths=None, maxdepth=None, bins=50, filename=None):
+    def animate_vertical_distribution(self, depths=None, maxdepth=None, bins=50, filename=None, subsamplingstep=1):
         """Function to animate vertical distribution of particles"""
         import matplotlib.pyplot as plt
         import matplotlib.animation as animation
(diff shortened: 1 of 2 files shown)


opendrift/models/oceandrift.py
```diff
@@ -435,7 +435,7 @@ def vertical_mixing(self, store_depths=False):
         else:
             return None
 
-    def animate_vertical_distribution(self, depths=None, maxdepth=None, bins=50, filename=None):
+    def animate_vertical_distribution(self, depths=None, maxdepth=None, bins=50, filename=None, subsamplingstep=1):
```
gauteh, line 438: Looks very useful, it would be very good with some documentation in this docstring about what the parameter does. See e.g. `def  plot` for a messy example :D

knutfrode:
Does this provide the same output if `subsmplingstep` is 1 (or not provided)?

manuelaghito:
Yes it should…. I couldn’t see differences but there was also a bug I
think, as explained in the commit description, which caused some problems a
in the oil vertical mixing example, there you should see some differences

tir. 16. nov. 2021 kl. 19:21 skrev Knut-Frode Dagestad <
***@***.***>:

knutfrode:
Very good

gauteh:
I still think it would be good to add some docs, otherwise it will be difficult for new users to discover and use this feature. Suggested:

```
Args:

     subsamplingstep (int): only plot `subsamplingstep` frames, use speed up animation time.
```

knutfrode:
Yes, docs are good.
But I guess this commit should give faster animation even if subsamplingstep is 1?

manuelaghito:
yes it is a bit faster even with subsamplingstep=1, but it was still quite slow in some cases. I can add some docs

knutfrode:
I think all OpenDrift animations can be speeded up a lot with `blitting`, especially if a landmask is involved.
I started on the main basemodel.animation(), but there it was a little involved, so have to continue en vakker dag.

But for this simpler method, I believe blitting should be very simple to implement.
https://matplotlib.org/stable/tutorials/advanced/blitting.html

PR #770, opened 2021-11-18
punkrockwriters

manuelaghito:
Improvements to animation generation and saving, and on plot_vertical_distribution()

With the default parameter fastwriter=False everything is still as before sho it should be safe to merge, but I think we can use fastwriter=True in all animation(), animation_profile() and animate_vertical_distribution(), or put it True directly in _save_animation(). I tested many examples and couldn't find problems

But nice if you also check first. 
logger.info call in _save_animation() shows elapsed time.
Check enclosed example


knutfrode:
Looks promising!
If this is fast and stable, it can soon replace the old saving method. No reason then to chose the slow method...
Perhaps @gauteh also can have a quick look before we merge?

gauteh:
Yes I'll take a look. If it does the same as previous I agree, we should
try hard to reduce unnecessary functionality and complexity.

fre. 19. nov. 2021, 08:48 skrev Knut-Frode Dagestad <
***@***.***>:

knutfrode:
One thing to keep in mind is that this must work also when running in batch, i.e. with no display/terminal available. But I guess this is what happens on CircleCI(?)

manuelaghito:
by the way.. trying to optimize animate_vertical_distribution() then I found out that most of the time (typically 80%) was used for _save_animation() then I googled around and found this solution, which really seems to help a lot both on gif and mp4

knutfrode:
Yes, very good. 
I guess that blitting is still an independent way to get even better performance. Have you tried that?
https://matplotlib.org/stable/tutorials/advanced/blitting.html

manuelaghito:
not sure that I did the blitting properly....but I tried different methods in animate_vertical_distribution() some with blit=True some with blit=false, (the alternative methods are commented and included in the code  - maybe this is not very clean/tidy ) but then I realized that the differences were small and the time used in saving was much larger.  

(Also... I could see that plot_vertical_distribution that basically do the same but doesn't generate a file is really fast compared to animate_vertical_distribution)

manuelaghito:
didn't tried blitting the way explained in the link

manuelaghito:
in https://stackoverflow.com/questions/30965355/speedup-matplotlib-animation-to-video-file

it was recommended to use this line for Matplotlib <3.4
self._frame_sink().write(self.fig.canvas.tostring_argb()) 

and replace it with this line for Matplotlib>=3.5
self._proc.stdin.write(self.fig.canvas.tostring_argb())

I just used the second option, which I guess is not supported in <3.4 do we need to worry about <3.4 ?

knutfrode:
Yes, we should now worry about Python<3.4
Merging now, will be interesting to see how it work!

manuelaghito:
now worry or not worry?

knutfrode:
not :-)

knutfrode:
be happy

knutfrode:
People with Python<3.4 should of course worry, but that is not our problem.

manuelaghito:
men det er matplotlib 3.5, (kke python 3.5) og jeg tror denne ble released kun i august 2021

knutfrode:
Beklager, tenkte det stod Python<3.4 (2014). Da bør vi kanskje legge inn en switch til den gamle varianten også.

manuelaghito:
so... now is fastwriter=True only used in example_oilspill_seafloor.py, so it shouldn't give a big improvement in the generation of all the figures in the gallery. In order to activate it by default everywhere then either we need to set fastwriter=True in the definition of the 3 animation/animte methods, or manually every time we call these methods, or we just remove all the referces to fastwriter in these methods and set it only i _save_animation(). What do you think is best? The last one is of course the simplest but less flexible.

knutfrode:
I guess we should do some intensive testing coming days/week, and then replace the old version with the new version as soon as it seems stable. Then we can remove the fastwriter-switch. There are already way too many input-parameters to the plot and animation methods.

manuelaghito:
matplotlib version..... wel well.... I actually have 3.4.1 and it's working for me :-)

manuelaghito:
ok i was confused, in the post he said <3.4 and >=3.4 so it is as expected

manuelaghito:
matplotlib 3.4 was releaseed in march 2021, still not too long ago. Do we need the switch?

knutfrode:
Yes, probably we then need it. I guess it is quite simple to add?

fre. 19. nov. 2021 kl. 14:30 skrev manuelaghito ***@***.***>:

PR #771, opened 2021-11-19
switch Matplotlib version punkrockwriters 

manuelaghito:
should be ready to merge.... althogh not testet for matplotlib < 3.4

### DIFF
--- a/examples/example_grid.py
+++ b/examples/example_grid.py
@@ -34,7 +34,7 @@ o.run(steps=60*2, time_step=1800, time_step_output=1800)
 #%%
 # Print and plot results
 print(o)
-o.animation(fast=False, corners=[3.5, 5.5, 59.9, 61.2])
+o.animation(fast=False, corners=[3.5, 5.5, 59.9, 61.2],fastwriter=True)
 
 #%%
 # .. image:: /gallery/animations/example_grid_0.gif
--- a/opendrift/export/punkrockwriters.py
+++ b/opendrift/export/punkrockwriters.py
@@ -10,6 +10,7 @@
 
 from matplotlib.animation import FFMpegWriter
 from matplotlib.animation import ImageMagickWriter
+import matplotlib
 
 class PunkImageMagickWriter(ImageMagickWriter):
     '''Faster ImageMagick-pipe writer bypassing figure.savefig.'''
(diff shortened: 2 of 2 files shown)


knutfrode:
Ser bra ut. Kan du også endre til `fastwriter=True` i `example_grid.py`?
Da kan vi se om det blir raskere enn nå på CircleCI (3min 11 sek):
https://opendrift.github.io/gallery/example_grid.html#sphx-glr-gallery-example-grid-py

manuelaghito:
3:11 min for å kjøre hele scripten?

55s versus 96s i min maskin

-

15:05:16 INFO    opendrift.models.basemodel: Saving animation to example_grid.gif...
15:05:16 INFO    opendrift.models.basemodel: Making animated gif...
15:06:11 INFO    opendrift.models.basemodel: fastwriter=True , elapsed_time: 55.161574286874384s

15:07:25 INFO    opendrift.models.basemodel: Saving animation to example_grid_slow.gif...
15:07:25 INFO    opendrift.models.basemodel: Making animated gif...
15:09:01 INFO    opendrift.models.basemodel: fastwriter=False , elapsed_time: 96.12291857996024s

manuelaghito:
image

knutfrode:
Det tar veldig mye lenger tid på CircleCI enn lokalt. Blir spennende å se nå straks hvor langt tid dette eksempelet tar med fastwriter.

manuelaghito:
image

manuelaghito:
158s in the gallery right now..... that's the interesting number.

knutfrode:
popcorn

knutfrode:
Actually not much difference in this case:
fastwriter

Thus something else must be the bottleneck on CircleCI.

knutfrode:
But the great speedup for local processing is of course a big plus anyway.

manuelaghito:
mmmmmmm... disappointing. Is CircleCI running many processes in parallel and allocating resources as needed? Could it be that since the other are slow that this single one doesn't need to be speeded up?

knutfrode:
Maybe it is because it is running in batch without display, and then it has to write to file anyway, somehow? I do not have enough insight in how renderers and these things work.

manuelaghito:
I mentioned yesterday (did I?) that when I use **%matplotlib inline** in ipython then there is no big advantage.... only when I use **%matplotlib qt** then I see the big advantage of 20-40%. I tried now again with %matplotlib inline, and then I get only 97s versus 105s..... this is more similar to CircleCI.. so maybe there is a common reason

manuelaghito:
when I run directy from terminal with python example_grid.py I get 87s/155s, much slower than in Spyder with **%matplotlib qt** this was also surprising.....

PR #782, opened 2021-11-29
Improved methods for saving animations

manuelaghito:
-new method based on saving and grab_frame method in matplotlib.animation writers

	the new method works significantly faster when MPLBACKEND=agg is used as for
	example for generation of sphinx gallery images in circleCI headless server

	the previous method ( baed on FuncAnimation ) is still used in interactive
	mode locally when no filename is given

-taking out the code that build filenames in sphinx gallery in separate method

-fix for comparing animations of different length (example_rungekutta_norkyst.py)


manuelaghito:
This should be a clean version equivalent to https://github.com/OpenDrift/opendrift/pull/775

manuelaghito:
image

check this out! 10m 32 s

manuelaghito:
what is missing to make "test_seed_letters" work? I thought you had fixed that already...

gauteh:
Maybe you need to rebase after getting latest master?

man. 29. nov. 2021 kl. 19:35 skrev manuelaghito ***@***.***>:

manuelaghito:
I don't understand why rebasing wouldn't work with [eba6e56](https://github.com/OpenDrift/opendrift/commit/eba6e56e0618c55e7f2683ee743e9ec6909f0caa), I did rebase before pushing, and I did it again, but it doesn't work ?!? So I added the fix again

manuelaghito:
ok da har jeg rebaset igjen da er det i orden. Kjørte docs-build-examples på nytt og i dag ble det 9m 20s :1st_place_medal:

manuelaghito:
this one should be ready to merge

knutfrode:
Seems good to me, if Gaute agrees?

gauteh:
I think there is a bit of debugging and formatting left over, otherwise it looks good. I could take a close look tomorrow with fresh eyes if you can wait until then ;)?

opendrift/models/basemodel.py
```diff
@@ -5188,65 +5251,68 @@ def add_halo_readers(self):
...
-                                'crop=trunc(iw/2)*2:trunc(ih/2)*2',  # cropping 1 pixel if not even
-                                '-pix_fmt',
-                                'yuv420p',
-                                '-an'
-                            ])
+                        FFwriter=animation.FFMpegWriter
                     elif fastwriter:
-                        FFwriter = PunkFFMpegWriter(
+                        FFwriter = PunkFFMpegWriter
+                    FFwriter = FFwriter(
```
gauteh, line 5315: This looks like a bug. Which probably also means that this code path is not tested.

knutfrode: which is a reminder that we should increase the general test coverage..... 53% is not very impressive https://coveralls.io/github/OpenDrift/opendrift?branch=master

manuelaghito: I don't think it's a bug, I tested it, it works... but maybe it's not elegant to define FFwriter first as a class and then as an instance? 

                    if not fastwriter:
                        **FFwriter**=animation.FFMpegWriter
                    elif fastwriter:
                        **FFwriter** = PunkFFMpegWriter
                    **FFwriter** = **FFwriter**(
                            fps=fps,
                            codec='libx264',
                            bitrate=1800,
                            extra_args=[
                                '-profile:v',
                                'baseline',
                                '-vf',
                                'crop=trunc(iw/2)*2:trunc(ih/2)*2',  # cropping 1 pixel if not even
                                '-pix_fmt',
                                'yuv420p',
                                '-an'
                            ])
                    anim.save(filename, writer=FFwriter)

manuelaghito: I have tested it it works, did it like this for not writing the same set of arguments two times

                    if not fastwriter:
                        FFwriter=animation.FFMpegWriter
                    elif fastwriter:
                        FFwriter = PunkFFMpegWriter
                    FFwriter = FFwriter(
                            fps=fps,
                            codec='libx264',
                            bitrate=1800,
                            extra_args=[
                                '-profile:v',
                                'baseline',
                                '-vf',
                                'crop=trunc(iw/2)*2:trunc(ih/2)*2',  # cropping 1 pixel if not even
                                '-pix_fmt',
                                'yuv420p',
                                '-an'
                            ])
                    anim.save(filename, writer=FFwriter)

knutfrode: ok, I believe this looks fine

knutfrode: ....though a lazy nitpicker might prefer
```
if fastwriter:
    FFwriter = PunkFFMpegWriter
else:
    FFwriter = animation.FFMpegWriter
```

manuelaghito:
what did you think about formatting Gaute, one line for each argument in method definitions?

gauteh:
I think the best is to just use `yapf` to autoformat. I will take a look on
code-style wrt to how you set up ifs and the class type, but you probably
have a pretty good way to do it. I think it is very important to add some
documentation about the arguments (maybe take in the suggestion I added in
the first punkrockwriter PR). Whether to use fast or fastwriter should
maybe also be documented in https://opendrift.github.io/performance.html .
Anyway, I'll take a closer look in not too long today.

tir. 30. nov. 2021 kl. 23:38 skrev manuelaghito ***@***.***>:

knutfrode:
If the fastwriter option is seen to be stable, I guess we can remove the `fastwriter` keyword and make it default?

gauteh:
Yes, agreed. Less complexity! I saw there were some conditionals on output type, does it support all formats?

gauteh:
First adding, and then removing an argument, is technically a breaking change so would be good to avoid that. Any users that started using it in the meantime would have their code broken if we remove it.

knutfrode:
Yes, so we should hurry to remove it before anyone notices....

gauteh:
yeah :) well, we haven't made any releases with it yet!

knutfrode:
Only formats supported are `mp4` and `gif`

gauteh:
Ok. So we need to keep an alternative method for other formats, and document that those two formats will probably be faster (probably give a logger.warning/info).

gauteh:
Does this PR replace #775?

manuelaghito:
I think I need to explain better here. In this version _save_animation() is actually never used unless hardcoded parameter force_FuncAnimation is set to True. _save_animation() is also the only function that uses fastwriter parameter. So there is actually quite a few lines of code that are not used at the moment. These methods were the best option before i found out the new even faster method used in _saving_animation(). 

Still, they work and might be a good option in case one want to work with Qt5agg and show interactive figures to screen and save files at the same time. In that case I see an advantage of using fastwriter=True, especially with mp4.

-

So, I didn't feel like removing this from the code, even if it is not used, and I didn't think it was a good idea to add even more parameters to animation() animation_profile() animate_vertical_distribution() to make it even more complicated.

(Also, since you decided to remove imagemagick dependency, i thought it was better to cut out this part of the code)

opendrift/models/basemodel.py
```diff
@@ -3732,22 +3732,52 @@ def plot_timestep(i):
...
-            logger.debug('Time to make animation: %s' %
-                         (datetime.now() - start_time))
-        else:
+
+        frames=x.shape[0]
+        if compare is not None:
+            frames=min(x.shape[0],cd['x_other'].shape[1])
+
+        mp4=False
+        if filename is not None: mp4 = str(filename)[-4:]=='.mp4'
```
gauteh, line 3741: Når du skal sjekke extension so er det betre å bruke `pathlib.Path.suffix`: https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.suffix. Slik det er no vil det krasje dersom filnavn har færre enn 4 chars.

manuelaghito: image

opendrift/models/basemodel.py
gauteh: Yeah, it should be changed to `else:`. Which puts me in the `lazy nitpicker` category ;)

opendrift/models/basemodel.py
gauteh: Du har rett, uansett so bør vi bruke python sin funksjonalitet for å sjekke extensions :)

manuelaghito:
> But it would be good if animation-method could make high-quality gifs with comparably small filesize in one go.

Also I must say that .gifs made with _saving_animation() use now PillowWriter and are saved very fast, but are larger files than those produced in earlier in _save_animation() with Imagemagick.

Maybe there are some additional parameters be to set properly, I don't know

opendrift/models/basemodel.py
```diff
@@ -5188,94 +5201,74 @@ def add_halo_readers(self):
...
 
         logger.debug(f"MPLBACKEND = {matplotlib.get_backend()}")
         logger.debug(f"DISPLAY = {os.environ.get('DISPLAY', 'None')}")
-        logger.debug(f"fastwriter: {fastwriter}")
+        logger.debug('Time to save animation: %s' %
+                     (datetime.now() - start_time))
 
-        if 'sphinx_gallery' in sys.modules:
-            plt.close()
+        plt.close()
```
gauteh, line 5271: Did you have to remove the 'sphinx_gallery' conditionals? Is always GUI plotting always done using FuncAnimation?

gauteh:
@manuelaghito: These errors seem legitimate: https://app.circleci.com/pipelines/github/OpenDrift/opendrift/2029/workflows/2400457c-7504-4546-b6ba-2f6cd91c0288/jobs/8218/tests

(index out of bounds)

did I break something? Or you?

manuelaghito:
I don't know? Everything worked before you fixed the code!!!? :-)

manuelaghito:
image

manuelaghito:
på linje 3921 skal det være x.shape[1] ikke x.shape[0]

manuelaghito:
:-1:  the code looks much nicer now, I guess the last two errors left on circleCI ar due to MET thredds server?

gauteh:
Yep, I think we are ready to merge!

gauteh:
We need a line for the changelog in history.rst by the way!

PR #793, opened 2021-12-03
Removing line with strange coordinates in io_netcdf.py

knutfrode:


knutfrode:
One failing example is due to inavailability of a Thredds datasett.

PR #798, opened 2021-12-13
Jul selafin reader

Boorhin:
A minor change of the selafin reader to be more logical in converting variable `ELEVATION Z` from the Selafin files
Now `ELEVATION Z` becomes `altitude` which is a standard name instead of artificially calling it `sea_floor_depth_below_sea_level`
the conversion into relative depth happens like before in the get_variable function by fetching the surface and the bottom of the mesh. no other change


knutfrode:
I assume this is now outdated.

PR #800, opened 2021-12-13
vertical_mixing: reflect particles from seafloor only if they are not supposed to be deactivated

manuelaghito:
Without this fix elements were being deactivated at the end of the mixing loop and "freezing" slightly above seafloor

### DIFF
--- a/examples/example_oil_ice.py
+++ b/examples/example_oil_ice.py
@@ -18,7 +18,7 @@ o.add_readers_from_list(['https://thredds.met.no/thredds/dodsC/barents25km_agg']
 #%%
 # Imaginary oil spill in Hinlopen strait
 o.seed_elements(lon=19.1909, lat=79.5986, radius=50,
-                number=3000, time=datetime.utcnow() - timedelta(hours=12))
+                number=3000, time=datetime.utcnow() - timedelta(days=7))
 
 #%%
 # Adjusting some configuration
--- a/opendrift/models/basemodel.py
+++ b/opendrift/models/basemodel.py
@@ -826,6 +826,7 @@ class OpenDriftSimulation(PhysicsMethods, Timeable):
         elif i == 'deactivate':
             self.deactivate_elements(self.elements.z < -sea_floor_depth,
                                      reason='seafloor')
+            self.elements.z[below] = -sea_floor_depth[below]
         elif i == 'previous':  # Go back to previous position (in water)
             logger.warning('%s elements hit seafloor, '
                            'moving back ' % len(below))
@@ -2406,6 +2407,7 @@ class OpenDriftSimulation(PhysicsMethods, Timeable):
         # Deactivate elements, if they have not already been deactivated
         self.elements.status[indices & (self.elements.status ==0)] = \
             reason_number
+        self.elements.moving[indices] = 0
         logger.debug('%s elements scheduled for deactivation (%s)' %
                      (np.sum(indices), reason))
         logger.debug(
--- a/opendrift/models/oceandrift.py
+++ b/opendrift/models/oceandrift.py
@@ -469,7 +469,7 @@ class OceanDrift(OpenDriftSimulation):
                 self.elements.z[reflect] = -self.elements.z[reflect]
 
             # Reflect elements going below seafloor
-            bottom = np.where(self.elements.z < Zmin)
+            bottom = np.where(np.logical_and(self.elements.z < Zmin, self.elements.moving == 1))
             if len(bottom[0]) > 0:
                 logger.debug('%s elements penetrated seafloor, lifting up' % len(bottom[0]))
                 self.elements.z[bottom] = 2*Zmin[bottom] - self.elements.z[bottom]


manuelaghito:
before
image

after
image

manuelaghito:
not so easy to see the difference in the example above since background diffusivity was small and 600m is quite deep. But still you can see it

In shallow waters with mixed layer going all the way down to see floor depth it would be more obvious, observing situations like this
image

manuelaghito:
but I guess we should discuss a bit about this, in my experiments I wanted to remove elements reaching seafloor by deactivating them, but probably it is a quite special case that in general is not very useful

knutfrode:
Yes, this depends on the module/application.
There are 3 options: relect, deactivate, or settle (for possible later resuspension).
This is supposed to be overloaded/subclassed by method `bottom_interaction()`

manuelaghito:
Ok, then there is probably a better way of solving this then the way I did, I see now that there is a "Let particles stick to bottom" section in vertical_mixing, but it didn't seem to work properly in my case. I don't know If I explained myself properly, but all those particles accumulating above seafloor in the screenshot shown above they are actually deactivated particles. But what I think happened was that they were deactivated only after the vertical_mixing loop, so they kept moving up and down during the vertical_mixing shorter time steps

            # Let particles stick to bottom
            bottom = np.where(self.elements.z < Zmin)
            if len(bottom[0]) > 0:
                logger.debug('%s elements reached seafloor, set to bottom' % len(bottom[0]))
                self.interact_with_seafloor()
                self.bottom_interaction(Zmin)

So they were not sticking to bottom here because they were already moved upwards in "  # Reflect elements going below "

So maybe this should be done earlier? Difficult to predict the consequences in all possible cases.

manuelaghito:
ok now I understand a little better what is going on.

some elements may be reflected from seafloor here, but then moved again below seafloor if they have some negative vertical velocity

            # Reflect elements going below seafloor
            bottom = np.where(self.elements.z < Zmin)
            if len(bottom[0]) > 0:
                logger.debug('%s elements penetrated seafloor, lifting up' % len(bottom[0]))
                self.elements.z[bottom] = 2*Zmin[bottom] - self.elements.z[bottom]

            # Advect due to buoyancy
            self.elements.z = self.elements.z + w*dt_mix*self.elements.moving

These are marked are scheduled for deactivation here in self.interact_with_seafloor()

            # Let particles stick to bottom
            bottom = np.where(self.elements.z < Zmin)
            if len(bottom[0]) > 0:
                logger.debug('%s elements reached seafloor, set to bottom' % len(bottom[0]))
                self.interact_with_seafloor()
                self.bottom_interaction(Zmin)

But deactivation is actually not executed before the whole vertical_mixing loop is completed, so in the mixing step they get reflected again.. and again... and again...

knutfrode:
It is generally a requirement that elements of any type shall not reach the surface or seafloor by mixing alone - as this is unphysical: the turbulent eddies do not penetrate surface or seafloor.
So sinking or rising is needed for this to happen, and therefor buoyancy is an integrated part of the vertical mixing loop, and cannot be performed before/after.

I have not looked at this in detail for a while: is the continued reflection of particles that are shceduled for deactivation only a (small) performance issue, or does it lead to wrong results as well?

manuelaghito:
Ok, it makes sense that elements that are affected by turbulence do not settle t seafloor.

-

I guess it's not a big deal but it does lead to wrong results..... it leads to the problem I was trying to address here and showing in the screenshots. 

If 'general:seafloor_action' is 'deactivate', and if elements sink, then you might get deactivated elements "frozen" slightly above seafloor. In m understanding this happens because the actual deactivation only occurs after the mixing loop is completed, while within the loop they are only scheduled for deactivation (is this correct?) and therefore still free to be reflected up

manuelaghito:
image

knutfrode:
Yes. I guess we could/should update the loop so that elements that are scheduled for deactivation during the loop, are not mixed further. If you agree?

manuelaghito:
yes, I can propose a fix for this

knutfrode:
Eccellente

manuelaghito:
example_oil_ice fails, is it because of thredds or because of previous merge or because of this commit?

knutfrode:
Great work! I guess we can keep using old data with oil-in-ice example.
Then ready to merge?

manuelaghito:
yes I guess it is safer to use older data, maybe happening again. It worked locally, so let's see if it works on circleci. Then I guess it should be ready to merge if you agree

manuelaghito:
this was quite crazy :-) I tried locally with -47 hours and it fails, -48 hours and it is ok! But on circleci they have a different time zone, so there it fails with -48 hours !!!

knutfrode:
Hehe, yes therefore the examples should rather use `utcnow()` than `now()`

manuelaghito:
don't understand why this keeps failing on the server, even with 36 hours.....and the example actually uses utcnow()....

manuelaghito:
Ok I was wrong thinking that it worked locally, with 48 hours it didn't crash immediately, but it still could complete the simulation.... so maybe now......

manuelaghito:
I think finally we are ready to merge

PR #836, opened 2022-01-20
Added methods to calculate DARPA and Liu-Weissberg skillscores in phy…

knutfrode:
…sics_methods. With unit-test, and demonstration in example_wind_drift_coefficient_from_trajectory.py


knutfrode:
Failing test is due to unavailable thredds dataset

PR #890, opened 2022-03-23
Added option to chose ensemble member in reader_netCDF_CF_generic.py

mateuszmatu:
Added functionality to the reader to chose which ensemble member should be used when working with ensemble data. 
Added an argument in the function call called "ensemble_member", which defaults to None. This argument takes in an integer. 
Added a if statement when initially reading in the data, which checks if the "ensemble_member" argument has been used, and if it has, then it will read in data from that specific ensemble member using xarrays .isel() option.

### DIFF
--- a/opendrift/readers/reader_netCDF_CF_generic.py
+++ b/opendrift/readers/reader_netCDF_CF_generic.py
@@ -114,8 +114,7 @@ class Reader(StructuredReader, BaseReader):
 
     """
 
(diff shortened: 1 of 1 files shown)


knutfrode:
Looks fine. Just to note that `ensemble_member` might be an in-house naming of the ensemble dimension, so might not work for ensemble files from third parties. But this can eventually be generalized later.

Small note: it is better in Python to use `is None` than `== None`

PR #894, opened 2022-03-24
perf(reader): replace the scipy.spatial.cKDTree with pygeos.STRtree to find the nearest node and element

### LINKED ISSUE #889: The difference between scipy.spatial.cKDTree and pygeos.STRtree
In the opendrift/readers/basereader/unstructured.py ，I found that _build_ckdtree_ and _nearest_ckdtree_ functions take a relatively long time to run the model, so I made the following changes in the method which find the nearest point:

```python
# ----------------------------unstructured.py----------------------------
# def _build_ckdtree_(self, x, y):
#     from scipy.spatial import cKDTree
#     P = np.vstack((x, y)).T
#     return cKDTree(P)
# Replace _build_ckdtree_ with _build_strtree_
def _build_strtree_(self, x, y):
    from pygeos import STRtree, points
    return STRtree(points(x,y))

#def __nearest_ckdtree__(self, idx, x, y):
#    q = np.vstack((x, y)).T
#    return idx.query(q, k = 1, workers = self.PARALLEL_WORKERS)[1]
# Replace __nearest_ckdtree__ with __nearest_strtree__
def __nearest_strtree__(self, idx, x, y):
    from pygeos import points
    return idx.nearest(pygeos.points(x,y))[1]
```
```python
# ----------------------------reader_netCDF_CF_unstructured.py----------------------------
# line 173
# self.nodes_idx = self._build_ckdtree_(self.x, self.y)
self.nodes_idx = self.__nearest_strtree__(self.x, self.y)

# line 176
# self.nodes_idx = self._build_ckdtree_(self.x, self.y)
self.faces_idx = self.__nearest_strtree__(self.x, self.y)
```

Here I use the flow field data of FVCOM and the normal wind field to drive openoil, so I only modified the reader_netCDF_CF_unstructured.py and unstructured.py, but the final graph is different from before the modification. I wonder if this modification is correct? If it's correct, then switching to the pygeos.STRtree can increase the operating speed of opendrift by more than 10%.

image

lyingTree:
Using pygeos.STRtree method will be more efficient than scipy.spatial.cKDTree when searching for the index of the nearest node and element.

Closes #889

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -29,6 +29,7 @@ dependencies:
   - coloredlogs
   - cmocean
   - pip
+  - pygeos
   - pip:
     - motuclient
     - roaring-landmask
--- a/opendrift/readers/basereader/unstructured.py
+++ b/opendrift/readers/basereader/unstructured.py
@@ -159,12 +159,23 @@ class UnstructuredReader(Variables):
         P = np.vstack((x, y)).T
         return cKDTree(P)
 
+    def _build_strtree_(self, x, y):
+        from pygeos import STRtree, points
+        return STRtree(points(x, y))
+
     def __nearest_ckdtree__(self, idx, x, y):
         """
         Return index of nearest point in cKDTree
         """
         q = np.vstack((x, y)).T
         return idx.query(q, k = 1, workers = self.PARALLEL_WORKERS)[1]
+
+    def __nearest_strtree__(self, idx, x, y):
+        """
+        Return index of nearest point in STRtree
+        """
+        from pygeos import points
+        return idx.nearest(points(x, y))[1]
 
     @staticmethod
     def __nearest_rtree__(idx, x, y):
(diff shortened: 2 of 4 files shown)


gauteh:
Thank you, have you tried to use strtree from shapely rather than from pygeos? Then we would avoid a new dependency, and pygeos will be deprecated in the future.

lyingTree:
> Thank you, have you tried to use strtree from shapely rather than from pygeos? Then we would avoid a new dependency, and pygeos will be deprecated in the future.

I have tried to use shapely to complete this function, but now it's not a good choice. 
I tried the shapely in 1.7.1, the main release version, and the  1.8.1.post1, the newest release version, but it's all unsatisfactory. The return in the shapely.strtree.STRtree.nearest is a geometry, and it can't return multi-value. Thus we have to reprocess this, which seriously affects the speed.

All in all, it's not a good choice now for here.
But for the Shapely 2.0, we should keep paying continuous attention.

PR #924, opened 2022-05-24
animation legend modifications

manuelaghito:
-fixing a small bug, white line around legend markers when rendering animation with fast=True
-new functionality for manually deactivating/ignoring some specific legend labels

### DIFF
--- a/opendrift/models/basemodel.py
+++ b/opendrift/models/basemodel.py
@@ -3651,6 +3651,8 @@ class OpenDriftSimulation(PhysicsMethods, Timeable):
                         [0], [0],
                         marker='o',
                         color='w',
+                        linewidth=0,
+                        markeredgewidth=0,
                         markerfacecolor=cmap(legend_index / (len(legend) - 1)),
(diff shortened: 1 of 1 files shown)


manuelaghito:
the artifact can be seen for example in https://opendrift.github.io/gallery/example_ensemble.html

knutfrode:
Jolly good. Ready to merge?

manuelaghito:
not sure if e2e0267 is problem free please have a look. This was a specific need I had with ChemicalDrift.... I wanted to eliminate some Chemical specis from the animation because they were not used... but I didn't want to change the colors of the remaining species.... this worked for me, for example setting legend=['dissolved', '', 'POC', 'sediment', ''] to ignore the second and fourth specie

manuelaghito:
I guess there if there are not known cases where we intentionally use empty strings as legend labels then this shoud not cause problems

manuelaghito:
added same trick for animation_profile. Ready to merge for me if all tests passed

PR #982, opened 2022-09-22
a couple of bugs in openoil's update_terminal_velocity()

GilesFearon:
Hi there, please consider these changes - I think they may be important for subsea releases


knutfrode:
Indeed, these seem to be two errors!
In addition, the same error of using diameter instead of radius applies to the high Reynolds regime. I will correct that in a next commit, and also update the tests to pass with these changes.
This seem to lead to some differences in vertical distribution of oil, but fortunately not as dramatic as one might fear.

PR #993, opened 2022-10-04
Speeding up allocation of self.history in import_file

manuelaghito:
for some reason numpy.ma.array allocation is extremely slow when mask=True is passed

### DIFF
--- a/opendrift/export/io_netcdf.py
+++ b/opendrift/export/io_netcdf.py
@@ -325,7 +325,8 @@ def import_file(self, filename, times=None, elements=None, load_history=True):
     if load_history is True:
         self.history = np.ma.array(
             np.zeros([num_elements, self.steps_output]),
-            dtype=history_dtype, mask=[True])
+            dtype=history_dtype)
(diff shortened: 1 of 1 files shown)


knutfrode:
Ah, yes!
In fact, we should aim to remove masked arrays completely from OpenDrift.

manuelaghito:
now it should be ok, hopefully :-) was very eager when I found the problem :-)
image

manuelaghito:
by the way if we don't need all values to be masked then we can remove the line **self.history[:] = np.ma.masked** then it is even faster. But at least like it is proposed now the initially allocated **self.history** should be exactly the same as before, as far as I can see, all zeros, all masked

manuelaghito:
just notied that I didn't pull the latest master with the recent commit https://github.com/OpenDrift/opendrift/commit/1d1df93aad71d111f313af34f763001e3bf7ef86

manuelaghito:
I am not sure what is going on here, I just close this and make a new one

PR #1011, opened 2022-11-01
Fix initial particle positions when using radius

simonweppe:
adding line below 

dist = dist * (radius/np.max(dist)) # set dist to range [0,radius]

np.random.randn(np.sum(number)) * radius return random numbers with mean=0 and variance=1 but they are not in range [-1,1].


knutfrode:
Documentation is here well hidden, but by default, radius is interpreted as a standard deviation and not abolute max, which makes sense for e.g. search and rescue where a initial position is known, but with some uncertainty.

If you want radius to be the max possible distance, and uniform spread within, you can add `radius_type='uniform'` to `seed_elements(....)`

simonweppe:
Ok my bad ! thanks for the clarification

PR #1038, opened 2022-12-12
correction factor to make diffusivity profile approach background diffusivity at depth==MLD

manuelaghito:
and avoid discontinuity at depth==MLD
should be ready to merge if you guys agree

this is just a small linearly increasing correction factor, which is 0 at depth=0, and gradually increasing to background_diffusivity when depth==MLD.

This discontinuity caused problems in some cases


manuelaghito:
before
image
after
image

manuelaghito:
check also the problem that I observed, now fixed
image
you can test this with axample_mixed_layer_depth.py, but you have to set the wind down to 5

knutfrode:
Very good!

Three tests are failing due to very slightly changed depths. You may adjust the values in the tests, and I can merge when tests are passing.

PR #1040, opened 2022-12-13
ChemD+RadioN: Bugfix. '.seconds' to to '.total_seconds()'

manuelaghito:
'.seconds' does not work for time_step >= 24 hours. Changed to '.total_seconds()'
@magnesim 
ready to merge I think


knutfrode:
Yes, `total_seconds` seems more correct here.
Tests are probably failing for some other reason.

knutfrode:
That said, `time_step` > 24 hours might not be a good idea?
`time_step_output` > 24h should however be ok.

manuelaghito:
I "needed" that to run some unit tests trying to make them fast. Dynamic partitioning seems to be a slow process.
time_step == 24 hours failed with '.seconds'... that gives 0!

PR #1041, opened 2022-12-13
ChemD: unit testing - dynamic partitioning

manuelaghito:
first ChemicalDrift test.
5s. Too slow? 
Seems to be impossible to make a ChemicalDrift simulation less then 3s.....
ready to merge


gauteh:
We have moved to the pytest framework for unit testing.

manuelaghito:
so, what I did is the old way? Is this ok or do I need to update it? I just copied what I found in  opendrift/tests/models/

manuelaghito:
not sure I understand the difference. Are test_models.py and test_stranding.py the old way?

knutfrode:
Yes, those are both old school, and should also be modernized one nice day.

Issue #713, opened 2021-09-28
issue with netcdf file export when using seed_cone with specified z, or other property, as array rather than single value 


simonweppe:
Hi all,

If one specifies an array rather than single value for one of the properties when seeding particle using `seed_cone()`, it seems that the netcdf export will fail because a "ndarray is not JSON serializable".

A fix could be to convert any properties that is an array to a string before saving to `self.seed_geojson` ? Not too sure what's a good practice for json in that context ?

For example below , included line 1638 of `basemodel.py. `
Happy to make a pull request if that's acceptable.

```
for key in properties.keys():
    if isinstance(properties[key],np.ndarray):
        properties[key] = np.array2string(properties[key])
```

PR #1074, opened 2023-03-28
Update basemodel.py

simonweppe:
Related to this : https://github.com/OpenDrift/opendrift/issues/713


knutfrode:
Seems good, thank you!

PR #1093, opened 2023-04-24
reader_netcdf: allow Dataset (in addition to filename) to be passed to reader

gauteh:


jerabaul29:
I am struggling to get this feature available by installing from git (see issue https://github.com/OpenDrift/opendrift/issues/1095 ). Would it be possible to push to conda the current state of opendrift so that this feature is available in the most up to date conda release? :)

PR #1096, opened 2023-04-26
Replaced fig.set_tight_layout(True) with matplotlib.rcParams['figure.…

knutfrode:
…autolayout'] = True

### DIFF
--- a/opendrift/models/basemodel.py
+++ b/opendrift/models/basemodel.py
@@ -21,31 +21,28 @@ import types
 import traceback
 import inspect
 import logging
-
 logging.captureWarnings(True)
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


gauteh:
Second option is better! Then we don't change global plotting settings for users just by importing opendrift.

knutfrode:
Yes, you are right.
Stikk fails, but will look at it later.

knutfrode:
It makes a minor difference if `fig.set_layout_engine('tight')` is executed before or after `fig.canvas.draw()` (one figure is slightly shifted vertically)
drawafter
drawfirst

Keeping it as it was (set_layout before canvas.draw), the tests pass, since figures are unchanged. This is probably ok.

PR #1105, opened 2023-05-02
Set same default ocean floor depth in SedimentDrift as in OceanDrift

poplarShift:
I just spent way too much time figuring out why my particles always ended up at -0.01 depth. Turns out the default depth of the SedimentDrift module is 0, which is not very helpful.

### DIFF
--- a/opendrift/models/sedimentdrift.py
+++ b/opendrift/models/sedimentdrift.py
@@ -54,7 +54,7 @@ class SedimentDrift(OceanDrift):
         'land_binary_mask': {'fallback': None},
         'ocean_vertical_diffusivity': {'fallback': 0.02},
         'ocean_mixed_layer_thickness': {'fallback': 50},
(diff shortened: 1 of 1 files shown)


knutfrode:
Yes, fallback of 0m ocean depth was not very meaningful. Though depth is supposed to be provided by an ocean model. Fallback of 10000m might be an ok workaround for modules where buoyancy is positive, and particles never settle. But for SedimentDrift I guessing having no real seafloor is normally not a good option?
Anyway, I will merge this update as soon as all tests are passing.

poplarShift:
Yes, the ocean model should of course provide it, so maybe throwing an error would be another way of handling it. Infinite (or very large) depth is useful only for debugging and quick test runs I suppose.

knutfrode:
An error is thrown for missing ocean depth if fallback value is set to None. But having 10000m allows running concept tests without real ocean depth. So can keep it like this.

PR #1122, opened 2023-05-25
Fix creation of concentraiton output file

magnesim:
Read lat and lon using get_variables function,
Set new max and min locations


knutfrode:
Very good, merging now.

magnesim:
The call of the 'write_netcdf_radionuclide_density_map' function could be commented, to avoid it from running in the daily updates.

PR #1140, opened 2023-07-10
Adding normal and lognormal droplet size distributions for subsea blowout

GilesFearon:
Added options to specify either a normal or lognormal initial droplet size distribution for a subsea blowout, which you may wish to consider including? I found them useful in a recent study where I used OpenDrift for some subsea blowout scenarios.

Also, there is a commit which I think addresses a small bug with the subsea blowout example where the droplet size range needs to be specified before seeding otherwise the specified range is not used (1faf1548465ad82b24cc9907ab1f8d2d98520d40).


knutfrode:
Thank you for this useful contribution.
Also I agree on the bug in the example.

PR #1179, opened 2023-10-30
OpenOil.seed_cone complains that mode is Ready and not config. Made h…

knutfrode:
…ack by changing to Config when setting oil_type config, and then back to Ready


knutfrode:
@gauteh I am making some adaptations to make examples pass. E.g. explicitly changing mode to Result in order to make plots without needing to run simulation first. Not sure if these types of changes seem healthy, or a bad sign?

gauteh:
If plot works without being in result mode then we can remove the requirement. It is almost always a bad sign to have to change the mode directly.

opendrift/models/openoil/openoil.py
```diff
@@ -1646,7 +1647,14 @@ def seed_elements(self, *args, **kwargs):
 
         if 'oil_type' in kwargs:
             if self.get_config('seed:oil_type') != kwargs['oil_type']:
-                self.set_config('seed:oil_type', kwargs['oil_type'])
+                if self.mode != Mode.Config:
```
gauteh, line 1650: Doesn't this already work as expected? It is not allowed to change the oil-type on later seed's? I.e. you can seed multiple times, but you can only use one oil_type.

opendrift/export/io_netcdf.py
```diff
@@ -261,6 +261,8 @@ def import_file_xarray(self, filename, chunks, elements=None):
         self.lonmax = np.float32(self.ds.lon.maxval)
         self.latmax = np.float32(self.ds.lat.maxval)
 
+    self.mode = Mode.Result
```
gauteh, line 264: Why do you need this? self.mode should never be modified directly, only through the `@required_mode` decorator.

examples/example_seed_demonstration.py
```diff
@@ -31,17 +22,20 @@
...
 #%%
 # Seeding 100 elements within a radius of 1000 m
+o = OceanDrift(loglevel=50)
 print('\n' + '='*70)
 print('Seeding 100 elements within a radius of 1000 m:')
 print('o.seed_elements(lon=4, lat=60, number=100, radius=1000, time=time)')
 print('='*70)
 o.seed_elements(lon=4, lat=60, number=100, radius=1000, time=time)
 #o.run(steps=1)
+o.mode = Mode.Result
```
gauteh, line 38: Better to remove the `@require_mode` decorator on `def plot` than setting the mode.

knutfrode: Yes, I now removed the requirement on `plot` and `animation` as these are made to also visualize seeding prior to the run

opendrift/export/io_netcdf.py
knutfrode: But in this case the simulation is imported from netCDF with Xarray, and data is fed into a new simulation object that is created. Thus it should be necessary in this case to explicitly set mode to Result?

gauteh: Yes, good point.

opendrift/models/openoil/openoil.py
knutfrode: I believe in fact this is a challenge for any property that can be provided to seed_command, but otherwise selected from default config.

knutfrode: No in fact I believe this is only a challenge for `oil_type` and Leeway `object_type`, as other element properties are allowed to be different from different elements. And config value is here not overwritten by seed value, just serving as a backup value.

gauteh: I think we can fix this by changing the mode after the seed call. But we have to be careful, because we must make sure the environment is finalized. I can take a look.

knutfrode: Yes, probably better than storing an internal flag `oil_type_already_chosen`
Though, repeated seed calls must be allowed, so changing mode might be problematic.

gauteh: That would allow it, but only with the same oil type.

gauteh: I added a __set_seed_config__ method that requires some special care (by us). We have to guarantee ourselves that it doesn't change the environment for instance.

Issue #1187, opened 2023-11-12
Enhancement idea:  Support reading from an Intake catalog!


rsignell:
I found out about OpenDrift last week while visiting @ruoying He's modeling group at North Carolina State University.  They were struggling with performance using the native ROMS reader with `open_mfdataset()` on a large collection of NetCDF files.   

I think it would be cool to build a reader that reads ROMS files (and potentially other types of files) into Xarray using an [Intake](https://intake.readthedocs.io/en/latest/) catalog.   This would allow folks to then customize the characteristics of the dataset for reading in OpenDrift simply by modifying the catalog entry.    Also there are a growing number of zarr format datasets and collections of NetCDF files that have been transformed by [kerchunk](https://fsspec.github.io/kerchunk/) into virtual datasets readable by the Zarr library, and these would be accessible as well!

gauteh:
Sounds great. It is possible to pass a xarray.Dataset to the generic reader already, this enables zarr datasets and possibly intake as well?

rsignell:
Oh, perhaps we need to look at that.  Our first thought was to modify the roms native reader, since the NCSTATE use case was with ROMS output.  We took a stab here: https://github.com/OpenDrift/opendrift/pull/1188  but don't really know what we are doing. 

But the example dataset we are using there that doesn't need any credentials to pull data via Intake from a collection of NetCDF3-64-bit offset files on S3-compatible object storage, so it might be nice use case to try out.

PR #1188, opened 2023-11-12
ROMS native intake reader

rsignell:
This is an attempt to add an Intake reader for ROMS, as raised in #1187.   It's not working yet. 

I have a [test notebook](https://nbviewer.org/gist/rsignell-usgs/8cd9e4d0fabd3cc8c508a400a3d11324) but it's bombing out on calculating the depths. 

The OpenDrift community might find the example dataset we are using interesting/useful:

* curvilinear ROMS grid for East Coast and Gulf Coast of US
* pulls data via Intake (using zarr library and kerchunk json file) to read from a collection of NetCDF3-64-bit offset files on S3-compatible object storage
* no `open_mfdatset` required
* no credentials are required!

``` python
import intake   
intake_catalog = 'https://mghp.osn.xsede.org/rsignellbucket1/rsignell/testing/cnaps.yml'
cat = intake.open_catalog(intake_catalog)
dataset = 'CNAPS_opendrift'
ds = cat[dataset].to_dask()  # creates Xarray dataset
```
to execute the above code you need `intake-xarray`, `zarr` and `s3fs` in your environment

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -2,16 +2,30 @@ name: opendrift
 channels:
   - conda-forge
 dependencies:
-  - matplotlib>=3.5
-  - numpy>=1.17
-  - scipy>=1.6
+  - matplotlib
+  - numpy
+  - scipy
   - netcdf4<=1.6.1
   - ffmpeg
-  - pyproj>=2.3
-  - libgdal>=3.1
-  - gdal>=3.1
+  - ipykernel
+  - pyproj
+  - libgdal
+  - gdal
   - xarray
-  - dask
+  - distributed
+  - intake-xarray
+  - cftime
+  - cf_xarray
+  - ipykernel
+  - ipywidgets
+  - panel
+  - fsspec
+  - pooch
+  - s3fs
+  - zarr
+  - hvplot
+  - geoviews
+  - datashader
   - cfgrib
   - pygrib
   - xhistogram
@@ -20,18 +34,20 @@ dependencies:
   - pytest-cov
   - pytest-benchmark
   - pytest-mpl
-  - cartopy>=0.20
+  - cartopy
   - nc-time-axis
   - geojson
-  - pynucos>=2.12
+  - pynucos
   - isodate
   - coloredlogs
   - cmocean
   - utm
-  - roaring-landmask>=0.7
-  - trajan>=0.1.3
+  - roaring-landmask
+  - trajan
+  - fastparquet
+  - nodejs
+  - coiled
   - adios_db
   - pip
   - pip:
-    - motuclient
-    - pytest-sphinx
+    - git+https://github.com/OpenDrift/opendrift@master
--- a/examples/example_roms_intake.py
+++ b/examples/example_roms_intake.py
@@ -1,0 +1,46 @@
+#!/usr/bin/env python
+"""
+ROMS 
+Intake reader
+==================================
+"""
+import numpy as np
+from opendrift.readers import reader_ROMS_intake
+from opendrift.models.oceandrift import OceanDrift
+import intake
+import xarray as xr
+from datetime import timedelta
+
+#from dask.distributed import Client
+#client = Client()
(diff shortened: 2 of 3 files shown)


rsignell:
@gauteh any chance you could take a quick look here? 
Does this part of the error with `Index is out of bounds for axis 0 with size 57` give any clues? 
There are 57 time steps in this test case...
``` 
12:25:33 INFO    opendrift.models.basemodel:1816: Duration, steps or end time not specified, running until end of first reader: 2023-11-03 00:00:00
12:25:33 INFO    opendrift.models.basemodel:891: Using existing reader for land_binary_mask
12:25:34 INFO    opendrift.models.basemodel:903: All points are in ocean
12:25:34 INFO    opendrift.models.basemodel.environment:670: ========================
12:25:34 ERROR   opendrift.models.basemodel.environment:671: Index is out of bounds for axis 0 with size 57
  File "/home/conda/global/e1003848400f1afa7580bb6fa3874f48f20b3522458965117a29babee3516098-20231112-153522-163623-16-opendrift/lib/python3.11/site-packages/opendrift/models/basemodel/environment.py", line 646, in get_environment
  File "/home/conda/global/e1003848400f1afa7580bb6fa3874f48f20b3522458965117a29babee3516098-20231112-153522-163623-16-opendrift/lib/python3.11/site-packages/opendrift/readers/basereader/variables.py", line 889, in get_variables_interpolated
    env, env_profiles = self.get_variables_interpolated_xy(
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

jbzambon:
+1 for this!

knutfrode:
I am not familiar with Intake, but 57 must be the length of the time dimension. So perhaps some dimensions got swapped somewhere?
Line 646, in `environment.py` indicates that lon/lat arrays have length equal to the time dimension (57) instead of the number of particles.
https://github.com/OpenDrift/opendrift/blob/master/opendrift/models/basemodel/environment.py#L646

rsignell:
Finally bit the bullet and learned how to use the debugging capabilities of VS Code to figure out what was going on here.   Turns out that there was nothing wrong with the PR.   The problem was with the Kerchunked dataset we were using to test it!   When you use Kerchunk  to aggregate along the time dimension, you need to supply an `identical_dims=[]` argument that specifies the variables you *don't* want to be aggregated along the time dimension.   And `Vtransform` was missing from that list.    So it had a time dimension and 57 values instead of 1.   Once that was fixed, the example `example_roms_intake.py` script works just fine! 
opendrift_intake

rsignell:
@knutfrode this is working and doesn't impact anything else in opendrift.    Please let me know what additional things I need to do -- I'd love to see this merged!

gauteh:
Hi, thank you for the pull request. We are not familiar with intake, but it seems like this could be useful for many? 

In general we prefer to use the `generic` reader as it is the most general and updated one. If this needs to be specific to ROMS it is better to subclass the existing ROMS reader to avoid duplication. One approach we have used with the generic reader is to allow xarray.Dataset's to be passed in, then you can set it up in advance using intake and the reader does not need to know about intake at all.

We are also very careful about adding dependencies. It looks like some of those are for cloud based reading, which I think is something that might happen anyway. Please reduce the additional dependencies to the absolute minimum.

If we can change the reader to take an xarray object, without adding dependencies I think that would be the best option. Is that possible?

gauteh:
That sounds great!

rsignell:
@gauteh yes that's a much better solution!

knutfrode:
This can be done the same way for the ROMS reader as for the generic netCDF Reader:
https://github.com/OpenDrift/opendrift/blob/master/opendrift/readers/reader_netCDF_CF_generic.py#L72

You are welcome to implement this and test on your own data, and then make a pull request after it works.

rsignell:
closing in favor of #1214

PR #1193, opened 2023-11-30
Allow lazy loading zarr dataset via netcdf reader

ways:
I've modified two files to allow lazy_reader passing zarr auth info to netcdf reader. Please let me know if it breaks anything.


knutfrode:
Looks good, and all tests passing. Merging now.

PR #1194, opened 2023-12-01
Addition to pullrequest 1193

ways:
A small detail was missing in https://github.com/OpenDrift/opendrift/pull/1193

I created a new variable in LazyReader, to hold on to the dataset (self._dataset). This is due to a "hack" in netcdf class where the variable filename can be a string OR a dataset, so I'm not the only one to blame for this being a bit custom. Shall I include more comments in the code to make it understandable?


knutfrode:
Yes, it is probably a good idea to include a small comment and eventually a link about this

ways:
Comments added.

PR #1214, opened 2024-01-17
adding xarray dataset option to ROMS reader

rsignell:
Following the pattern of the generic CF reader, adding the option to pass a xarray dataset to the ROMS native reader. 

I wasn't sure if we wanted an example using this approach, because it's just a minor mod of the existing `example_roms_native.py` test. 

```
(odrift) rsignell@OSC:~/repos/opendrift/examples$ diff example_roms_native.py  example_roms_native_xarray_dataset.py
9a10
> import xarray as xr
14,15c15
< # Creating and adding reader for Nordic 4km current dataset
< nordic_native = reader_ROMS_native.Reader(o.test_data_folder() +
---
> filename = (o.test_data_folder() +
16a17,21
>
> # Creating and adding reader for Nordic 4km current dataset
>
> ds = xr.open_dataset(filename, decode_times=False)
> nordic_native = reader_ROMS_native.Reader(ds)
```


knutfrode:
Yes, this looks fine.
But I agree that the example might not be needed, as it does not add much more than illustrating that the reader can take Xarray datasets as an alternative to a filename or URL - which should now also be clear from the docstring.

It might be more interesting to show an example illustrating usage of intake - however, this would require four more dependencies (intake, intake-xarray, zarr, s3fs), and we would like to keep dependencies to a bare minimum. And generally it is probably better to avoid direct links between intake and OpenDrift, and instead let Xarray be the common meeting point.
But intake does look interesting, we will get familiar with it!

Thus, if you could remove the example from this PR, we can then merge it.

knutfrode:
Ok, maybe the simplest first step would be if rsignell could add these minimum dependencies to `environment.yml` and update the example script to using the intake catalog as in #1188 (assuming the test dataset will remain on the server for some time).
Then we could eventually later move these dependencies to an `extras`-file, if it should be needed.

rsignell:
I've removed the example for now, but will submit another PR to include an intake example soon -- we have a nice example that uses ROMS output on AWS Open Data, but we are waiting for a new release of fsspec that addresses a bug that turned up with that workflow.

knutfrode:
Yes, this is good. Merging now.

rsignell:
Thanks so much @knutfrode and @gauteh for bringing this one home!

Would it be too much to ask for a new tag so we can get a new version of OpenDrift on conda-forge?   😬

knutfrode:
We will most probably make a new release for Conda within a week or two

gauteh:
Hi, I have made a new release now.

rsignell:
@gauteh thanks so much!   I see the feedstock action is failing: https://github.com/conda-forge/opendrift-feedstock?tab=readme-ov-file#current-build-status    
Should I ask @ocefpaf for help?

gauteh:
No, no need to ask. If you see the logs it's just an outdated dependency. You can make a PR with updated deps.

PR #1225, opened 2024-02-12
changes to ROMS reader for wetdry masks

kthyng:
Main changes are to allow for the land_binary_mask as well as mask_rho, mask_u, and mask_v to have 3D or 2D dimensions since masks in ROMS can vary in time if they represented cells that can wet and dry in time. With the changes, if a variable `wetdry_mask_rho` is available, it will be used for the rho mask in place of `mask_rho`, otherwise will use `mask_rho`, and the same for the u and v masks. The masks for different grids use the same indices, same as before these changes, but now they are managed a little more simply in one location in the function. Additionally, more effort is made to match variables to the appropriate mask since sometimes they may not be a typical ROMS- output variable.

More changes:

* masks are saved in a dict in case they are reused across different variables
* the mask is applied to the variable array as a bool index directly to convert to nans. This is better than the previous approach of searching for the missing value and replacing that with nan because with wet/dry masks the dry cells are not always represented by a fill value and instead can only be found using the mask.
* added some more variables to check for whether to rotate or not and added log statements too.

### DIFF
--- a/opendrift/readers/reader_ROMS_native.py
+++ b/opendrift/readers/reader_ROMS_native.py
@@ -286,15 +286,78 @@ class Reader(BaseReader, StructuredReader):
         # Run constructor of parent Reader class
         super(Reader, self).__init__()
 
+    @property
+    def mask_rho(self):
+        """Mask for the rho-points.
+        
+        Uses wetdry_mask_rho (which should be 3D) if available, otherwise mask_rho (2D).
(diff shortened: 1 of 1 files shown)


kthyng:
@knutfrode here is that PR I mentioned for dealing with wet/dry masks in ROMS (along a few other items).

knutfrode:
Looks all good, and all tests and examples are passing, thus merging now.
Thank you for the useful improvement.

Issue #1130, opened 2023-06-13
Adding projection to ROMS native reader


tklenz:
Hi there, 

I am trying to calculate LCS's from a ROMS simulation and am running into a bit of an issue. Using the native reader "as is" the fakeproj correctly returns the corners of my domain, but only places one particle in it.

===========================
Reader: roms native
Projection:
  None
Coverage: [pixels]
  xmin: 0.000000   xmax: 793.000000   step: 1   numx: 794
  ymin: 0.000000   ymax: 361.000000   step: 1   numy: 362
  Corners (lon, lat):
    (199.67,  55.23)  (213.36,  63.39)
    (206.19,  52.26)  (220.02,  59.70)
[...]

09:27:29 INFO    opendrift.models.basemodel:1701: All points are in ocean
09:27:29 INFO    opendrift.models.basemodel:2870: 2016-08-02 16:00:00 - step 1 of 120 - 1 active elements (0 deactivated)
09:27:29 INFO    opendrift.models.basemodel:2870: 2016-08-02 15:50:00 - step 2 of 120 - 1 active elements (0 deactivated)

Consequently, no ftles can be calculated. From a previous thread (issue #783) I gathered that I need to add a projection to the reader and did so in opendrift.readers.basereader.variables line 14:

class ReaderDomain(Timeable):
    """
    Projection, spatial and temporal domain of reader.
    """
    name = None

    proj4 = '+proj=lcc +lat_1=55 +lat_2=60 +lon_0=-95'
    proj = None

Using print(reader) I get this: 

===========================
Reader: roms native
Projection:
  +proj=lcc +lat_1=55 +lat_2=60 +lon_0=265 +units=m
Coverage: [degrees]
  xmin: -3531750.000000   xmax: -2342250.000000   step: 1500   numx: 794
  ymin: -3003750.000000   ymax: -2462250.000000   step: 1500   numy: 362
  Corners (lon, lat):
    (-111.88, -15.26)  (-106.32, -14.12)
    (-111.27, -17.62)  (-105.91, -16.57)

and finally particles are placed on the grid

17:00:50 INFO    opendrift.models.basemodel:2762: Backwards simulation, starting at time of last seeded element
17:01:45 INFO    opendrift.models.basemodel:1689: Using existing reader for land_binary_mask
17:01:45 INFO    opendrift.models.basemodel:1701: All points are in ocean
[...]

PR #1228, opened 2024-02-14
Allow for saving of interpolators

kthyng:
I made changes to structured.py to allow for saving the interpolators to save time on repeated simulations. The necessary attributes to use the save option need to be input via the reader, currently only available in the ROMS reader.

This PR also has changes that clean up the name variable in the ROMS reader. Previously there was logic to deal with name for several cases but subsequently it was overwritten to be “roms native”. Now name is used if input and otherwise set to “roms native”.

The change to allow for saving interpolators was mentioned in https://github.com/OpenDrift/opendrift/issues/1130.

### DIFF
--- a/opendrift/readers/basereader/structured.py
+++ b/opendrift/readers/basereader/structured.py
@@ -1,7 +1,9 @@
+import pickle
 import numpy as np
 import pyproj
 from scipy.ndimage import map_coordinates
 from abc import abstractmethod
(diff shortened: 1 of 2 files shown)


kthyng:
Sorry I didn't notice there were updates I hadn't incorporated into my branch.

knutfrode:
Thank you, this is a useful feature.

PR #1237, opened 2024-02-29
fast and small animation files by selecting frames as range or list

manuelaghito:
this should be ready to merge.
this works also without selecting a subset of output frames with o.run, which is undesired in some cases


knutfrode:
Nice!

Issue #1235, opened 2024-02-26
Question: how does ROMS work for 3D without zeta included?


kthyng:
I've been using the ROMS reader a lot and some of it has remained opaque to me. One question is: how is depth handled given that zeta (sea surface height) is not used and zeta is required to calculated the depth of the vertical layers in time for ROMS? Is it assuming quiescent surface for calculating the vertical layers?

knutfrode:
I have in fact not used that reader much myself, but made the original version based on some local ROMS files which did not even contain zeta, as bathymetry is generally quite steep off the Norwegian coastline, so that zeta << H.

The sigma-to-z conversion is made using a computationally efficient method from the package Roppy of Bjørn Aadlandsvik (https://opendrift.github.io/_modules/opendrift/readers/roppy/depth.html#sdepth) where zeta is not taken into account.
If zeta exists in the file, it will be returned as a regular variable (along with currents, temperature etc), that the "module developer" can access and use e.g. within the `update` method.

However, I am not sure whether this is strictly correct, or if zeta should have been taken into consideration in the conversion from sigma to z, so that z would be 0 at the very surface (zeta)? If so, I guess zeta should again be subtracted within the reader to be consistent with the general convention in OpenDrift that z refers to mean sea level.
In other words, I guess the main question is whether zeta should be included in the stretching of vertical coordinates, or if it can be added afterwards?
I would assume any difference would be negligible for our local applications where bathymetry is steep off the coast (zeta<<H), but it might perhaps be important for your applications that even include wetting and drying?


(As perhaps already meantioned: it is  a future goal that the generic netCDF-reader should also support other vertical coordinates than z, so that a dedicated ROMS reader should not be needed. Though that might require ROMS output to be strictly adhering to CF-convention, which it was not at the time when the ROMS reader was orinally made.)

knutfrode:
Making the generic reader support ROMS native output would be a long term goal, but I believe that a general adaptation of OpenDrift for better support of water and particles above z=0 should hopefully be fairly straightforward to do on short time, although it is a quite fundamental change.

It seems we then agree that z=0 should still mean the mean sea level, and that water and particles can then have z-values between `seafloor_depth_below_mean_sea_level` (constant in time) and `zeta` (variable in time).

I can think of a few necessary adaptations:
[...]

PR #1238, opened 2024-02-29
Add zeta to sdepth in depths.py (part of effort to add zeta to ROMS reader)

kthyng:


knutfrode:
This looks very good, including the new unittests.
The slight modifications to the existing test seem also very reasonable.

Thus, I might suggest that this should be safe to merge as a clear improvement, before proceeding with other updates, where there are some other challenges as discussed here: https://github.com/OpenDrift/opendrift/issues/1235#issuecomment-1972196208

kthyng:
ok I added in my addition of zeta to the ROMS reader now too which was missing before.

knutfrode:
Ref https://github.com/OpenDrift/opendrift/issues/1235, I believe that this PR could be merged.

Some tests are failing, but this is presumably just do to minor changes in depths, and that tests could be updated accordingly.
So if you could adjust these tests to pass, I can merge this PR.

kthyng:
Thank you, I should be able to get back to this and other on-going development work on opendrift by mid-next week. In particular I have some changes to allow for the wetting/drying mask that follow up on these changes — should they be in a different PR?

knutfrode:
It could be good to merge this first, but you can decide if you prefer to add more before merging.

But CircleCI shows that one of the tests (test_truncate_ocean_model) is failing due to an index error in the ROMS reader:
https://app.circleci.com/pipelines/github/OpenDrift/opendrift/2869/workflows/02ca903c-7b93-4a2e-871c-a9bcc1849066/jobs/13691/parallel-runs/1/steps/1-110
Screenshot from 2024-03-14 19-13-45

The special config setting `o.set_config('drift:truncate_ocean_model_below_m', <depth>)` is available for users who want to increase performance by only retrieving the upper few layers of an ocean model, where the lowest layer is extrapolated towards seafloor for any particles below that level:
https://github.com/OpenDrift/opendrift/blob/master/tests/models/test_run_3d.py#L83
So apparently a small fix has to be made in the ROMS reader to account for this possibility.

knutfrode:
It turns out that it fails on the first run of `test_truncate_ocean_model`, before the truncation is applied.
Thus it seems to be a general (indexing) problem in the ROMS reader, not related to the optional truncating.
I can have a closer look at this soon.

kthyng:
I'm digging back into this work today so I'll see what I can figure out too.

knutfrode:
> I'm digging back into this work today so I'll see what I can figure out too.

Ok.
Btw, the depth-module from the Roppy package was copied into Opendrift a long time ago, but it has since evolved also in the Roppy repository:
https://github.com/bjornaa/roppy/commits/master/roppy/depth.py

Generally, I think it would be better if OpenDrift could instead use Roppy as an external dependency, instead of just copying this single file. Thus it would be good if your updates with zeta could be incorporated into Roppy official.
But @bjornaa should decide whether this is desirable, as any change in the Roppy API might have downstream consequences for others.

kthyng:
Yes I agree that would be a better path. I'd be happy to make changes to roppy directly if @bjornaa would be interested.

kthyng:
Ok how about this:
I need to do some clean up on my end to save some subsequent code changes I have and look at that test, but then what if we push the changes in this PR through now. Then, once we hear back from @bjornaa we can presumably subsequently export this capability to roppy instead of having it built into opendrift, but do it as a subsequent step.

knutfrode:
Yes, we should anyway document and test the new improvement in OpenDrift, before eventually suggesting the same update to roppy.
But before merging into OpenDrift all tests must be passing.

opendrift/readers/reader_ROMS_native.py
```diff
@@ -424,11 +446,13 @@ def get_variables(self, requested_variables, time=None,
...
             if self.z_rho_tot is None:
                 Htot = self.sea_floor_depth_below_sea_level
-                self.z_rho_tot = depth.sdepth(Htot, self.hc, self.Cs_r,
+                zeta = self.zeta[indxTime]
+                self.z_rho_tot = depth.sdepth(Htot, zeta, self.hc, self.Cs_r,
                                               Vtransform=self.Vtransform)
 
             H = self.sea_floor_depth_below_sea_level[indy, indx]
-            z_rho = depth.sdepth(H, self.hc, self.Cs_r,
+            zeta = self.zeta[imask]
```
knutfrode, line 452: This line seems to fail with the following error:
```
ERROR    opendrift.models.basemodel.environment:environment.py:674 index 52 is out of bounds for axis 0 with size 3
  File "/root/project/opendrift/models/basemodel/environment.py", line 649, in get_environment
  File "/root/project/opendrift/readers/basereader/variables.py", line 898, in get_variables_interpolated
    env, env_profiles = self.get_variables_interpolated_xy(
  File "/root/project/opendrift/readers/basereader/variables.py", line 764, in get_variables_interpolated_xy
    env, env_profiles = self._get_variables_interpolated_(
  File "/root/project/opendrift/readers/basereader/structured.py", line 278, in _get_variables_interpolated_
    self.get_variables(blockvariables_before, time_before,
  File "/root/project/opendrift/readers/reader_ROMS_native.py", line 454, in get_variables
    zeta = self.zeta[imask]
  File "/opt/conda/envs/opendrift/lib/python3.10/site-packages/xarray/core/variable.py", line 764, in __getitem__
    data = as_indexable(self._data)[indexer]
  File "/opt/conda/envs/opendrift/lib/python3.10/site-packages/xarray/core/indexing.py", line 700, in __getitem__
    return type(self)(_wrap_numpy_scalars(self.array[key]))
  File "/opt/conda/envs/opendrift/lib/python3.10/site-packages/xarray/core/indexing.py", line 667, in __getitem__
    return type(self)(_wrap_numpy_scalars(self.array[key]))
  File "/opt/conda/envs/opendrift/lib/python3.10/site-packages/xarray/core/indexing.py", line 567, in __getitem__
    return type(self)(self.array, self._updated_key(indexer))
  File "/opt/conda/envs/opendrift/lib/python3.10/site-packages/xarray/core/indexing.py", line 533, in _updated_key
    full_key.append(_index_indexer_1d(k, next(iter_new_key), size))
  File "/opt/conda/envs/opendrift/lib/python3.10/site-packages/xarray/core/indexing.py", line 285, in _index_indexer_1d
    indexer = _expand_slice(old_indexer, size)[applied_indexer]
IndexError: index 52 is out of bounds for axis 0 with size 3
```

kthyng:
I see that the tests are all failing here but it looks like they fail on build? They are running for me locally anyway.

knutfrode:
Tests are also passing for me with this PR.
Thus merging now.

PR #1251, opened 2024-03-12
state_to_buffer save

climberlenny:
there was a bug writing data to the netCDF file. In some rare situations, step 100 (the default) is skipped, leading to an error as the buffer_limit _length is exceeded in the next step.

The following code produce the error :
`from datetime import datetime, timedelta
from opendrift.models.oceandrift import OceanDrift

o = OceanDrift(loglevel=20)

o.set_config("general:use_auto_landmask", False)
o.set_config("environment:constant:land_binary_mask", 0)
o.set_config("drift:max_age_seconds", 100)
start_time = datetime(2024, 1, 1)

o.seed_elements(lon=3, lat=60, time=start_time)
o.seed_elements(lon=3, lat=60, time=start_time + timedelta(hours=1) * 7)
o.seed_elements(lon=3, lat=60, time=start_time + timedelta(hours=1) * 100)

o.run(
steps=200,
export_buffer_length=5,
outfile="test.nc",
time_step=3600,
time_step_output=3600 * 2,
)`

### DIFF
--- a/opendrift/models/basemodel/__init__.py
+++ b/opendrift/models/basemodel/__init__.py
@@ -1971,12 +1971,12 @@ class OpenDriftSimulation(PhysicsMethods, Timeable, Configurable):
 
                 if self.num_elements_active(
                 ) == 0 and self.num_elements_scheduled() > 0:
-                    self.steps_calculation += 1
                     logger.info(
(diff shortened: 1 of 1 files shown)


knutfrode:
Very good. I will merge, but then remove the inline comments, as the mentioned line numbers will be outdated and wrong after future updates.

PR #1259, opened 2024-03-20
IO module for parquet

poplarShift:
For long-running simulations with limited particle lifetimes, in-memory sizes of netcdf arrays become quite large. Also, (at least for me) the first operation I do with the netcdf files usually is `xr.open_dataset(...).to_dataframe()`.This is a first draft that writes directly to a tabular format.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -34,4 +34,3 @@ dependencies:
   - pip
   - pip:
     - motuclient
-    - pytest-sphinx
--- a/opendrift/export/io_parquet.py
+++ b/opendrift/export/io_parquet.py
@@ -1,0 +1,67 @@
+import sys
+import os
+from datetime import datetime, timedelta
+import logging
+
+logging.captureWarnings(True)
+logger = logging.getLogger(__name__)
+import string
+from shutil import move
+
+import numpy as np
+import pandas as pd
+from opendrift.models.basemodel import Mode
+
+def init(self, filename):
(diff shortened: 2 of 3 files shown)


knutfrode:
Good, nice to have the first alternative writer to the default netCDF-writer.
I did some modifications, with the help from @gauteh 

Note that fastparquet is made an optional dependency (not included in environment.yml), and the test is skipped unless fastparquet is installed explicitly by the user.
Now all tests are passing, so merging this PR.

poplarShift:
That was quicker than I expected, thanks! I believe there are still some unused imports that could have been removed - will make a separate PR for that then.

knutfrode:
Yes, I removed those imports as well, and simplified example a little. But did not mention this in the commit-comments.
We like tests to be as fast as possible, but not too critical here as it is only run for users who chose to install fastparquet.

knutfrode:
Hi,
with OpenDrift 1.13.0 comes some small API changes, generally for the better:
https://opendrift.github.io/history_link.html#release-v1-13-0

However, the parquet writer will also need some corresponding updates.
I expect this to be straightforward, basically obtaining information from Xarray Dataset `self.result` instead of from the removed `self.history / self.steps_output / self.steps_exported`.

Would you like to have a go yourself to update the writer, and eventually also update the unit-test to check that the produced file is correct?

Issue #1196, opened 2023-12-06
ROMS and wetting/drying masks


kthyng:
Hi! I need to override the constant values set in `consts.py` for "sea_floor_depth_below_sea_level" because in ROMS the depths are negative instead of positive and the default range defined is 0 to 12000. I have tried a few approaches for this, most recently:

Initial setup:
```
reader = reader_ROMS_native.Reader(loc)
ocean_drift_sim = OceanDrift(loglevel=20)
ocean_drift_sim.add_reader([reader])
```

then:
```
key = "sea_floor_depth_below_sea_level"
c = {}
c[f'environment:fallback:{key}'] = ocean_drift_sim.get_configspec()[f'environment:constant:{key}']
c[f'environment:fallback:{key}'].update({"min": -12000, "max": 0})
ocean_drift_sim._add_config(c)
ocean_drift_sim.set_config(f'environment:fallback:{key}', -10000)
```

Also, separately, tried this:
```
ocean_drift_sim.get_configspec()['environment:constant:sea_floor_depth_below_sea_level']["min"] = -12000
ocean_drift_sim.get_configspec()['environment:constant:sea_floor_depth_below_sea_level']["max"] = 0
```

But when I actually then seeded the simulation and ran it, I still got log messages like:
```
11:22:54 WARNING opendrift.readers.basereader.variables:635: Invalid values (-6.475353 to -0.3697905) found for sea_floor_depth_below_sea_level, replacing with NaN
11:22:54 WARNING opendrift.readers.basereader.variables:638: (allowed range: [0, 12000])
```

or in the case that I changed the fallback value:
```
16:26:17 WARNING opendrift.readers.basereader.variables:635: Invalid values (-10000.0 to -10000.0) found for sea_floor_depth_below_sea_level, replacing with NaN
16:26:17 WARNING opendrift.readers.basereader.variables:638: (allowed range: [0, 12000])
```

So, neither of these approaches seemed to "stick" and it seems that what is hard-wired into "consts.py" isn't meant to be overridden. Any suggestions for how to proceed? Thanks.

kthyng:
The ROMS reader does seem to think the depths should be negative, for example:

https://github.com/OpenDrift/opendrift/blob/15546beb529f0d94629dc971a00b6edcb80dbf2b/opendrift/readers/reader_ROMS_native.py#L74-L78

unless I am over-interpreting those lines as being more meaningful than they are.

kthyng:
[...]

PR #1264, opened 2024-03-28
Updates to ROMS zeta, seafloor checks, tests

kthyng:
* added sea_surface_height to OceanDrift and models that subclass OceanDrift, also a new parameter
* interact_with_seafloor accounts for zeta in bottom check, as does other bottom check in OceanDrift
* angle and zeta are updated in ROMS reader
* z_rho calculation is shifted so that it is 0 at the surface like z 
* tests are updated, including a ROMS reader specific test file
* changed sea_floor_depth_below_sea_level to -20 from -10 for wet/dry cases that are a little more extreme

### DIFF
--- a/opendrift/models/basemodel/__init__.py
+++ b/opendrift/models/basemodel/__init__.py
@@ -694,8 +694,25 @@ class OpenDriftSimulation(PhysicsMethods, Timeable, Configurable):
             return
         if 'sea_floor_depth_below_sea_level' not in self.env.priority_list:
             return
+
+        if not hasattr(self, 'environment') or not hasattr(
+                self.environment, 'sea_surface_height'):
+            logger.warning('Seafloor check not being run because sea_surface_height is missing. '
+                           'This will happen the first time the function is run but if it happens '
+                           'subsequently there is probably a problem.')
+            return
+
+        # the shape of these is different than the original arrays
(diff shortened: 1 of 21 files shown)


knutfrode:
Really a great job, thank you very much!
All tests are passing, and your code changes and additions look all good to me.

So I can merge this, if you think it is ready to go?

kthyng:
@knutfrode Thank you! I was stuck for awhile thinking between z_rho and z but when I realized I could just convert from z_rho to z with zeta without something more complicated I was trying, it all worked out ok.

I am feeling pretty good about it all at this point and I think it is ready to go. I think this closes https://github.com/OpenDrift/opendrift/issues/1196 (along with https://github.com/OpenDrift/opendrift/issues/1235) also unless I realize another thing I haven't noticed 😆

knutfrode:
Yes, we can make a new release within a few days.

knutfrode:
Release 1.11.2 is now out. I added some release notes here:
https://opendrift.github.io/history_link.html

Feel free to suggest other formulations for your contributions, e.g. by editing history.rst directly in your browser:
https://github.com/OpenDrift/opendrift/blob/master/history.rst

It should be available on conda-forge within a couple of hours:
https://anaconda.org/conda-forge/opendrift

PR #1265, opened 2024-03-28
small fix

kthyng:
assertion line for z_rho in ROMS reader does not ignore nans. But also it is probably too much to have that assertion so fixed to ignore nans but also changed to a warning.

### DIFF
--- a/examples/example_vertical_mixing.py
+++ b/examples/example_vertical_mixing.py
@@ -38,8 +38,9 @@ o.seed_elements(lon=4, lat=60, z=seed_depth, time=time, number=N, terminal_veloc
 o.time = time
 o.time_step = timedelta(hours=hours)
 o.release_elements()
-o.environment = np.array(np.ones(N)*sea_floor_depth,
-                dtype=[('sea_floor_depth_below_sea_level', np.float32)]).view(np.recarray)
+o.environment = np.array(np.ones(N, 0)*sea_floor_depth,
+                dtype=[('sea_floor_depth_below_sea_level', np.float32),
+                                ('sea_surface_height', np.float32)]).view(np.recarray)
 o.environment_profiles = {
         'z': z,
         'ocean_vertical_diffusivity':
--- a/examples/example_vertical_mixing.py
+++ b/examples/example_vertical_mixing.py
@@ -38,9 +38,9 @@ o.seed_elements(lon=4, lat=60, z=seed_depth, time=time, number=N, terminal_veloc
 o.time = time
 o.time_step = timedelta(hours=hours)
 o.release_elements()
-o.environment = np.array(np.ones(N, 0)*sea_floor_depth,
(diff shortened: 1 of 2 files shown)


kthyng:
@knutfrode sorry a fix already!

knutfrode:
All tests are passing, however, this example script is failing:
https://opendrift.github.io/gallery/example_vertical_mixing.html
```
    ../../examples/example_vertical_mixing.py failed leaving traceback:

      File "/root/project/examples/example_vertical_mixing.py", line 51, in <module>
        depths = o.vertical_mixing(store_depths=True)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/root/project/opendrift/models/oceandrift.py", line 502, in vertical_mixing
        Zmin = -1.*(self.environment.sea_floor_depth_below_sea_level + self.environment.sea_surface_height + Dcrit)
                                                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/opt/conda/envs/opendrift/lib/python3.11/site-packages/numpy/core/records.py", line 455, in __getattribute__
        raise AttributeError("recarray has no attribute %s" % attr) from e
    AttributeError: recarray has no attribute sea_surface_height
```

This is a bit special example, that runs the mixing loop without running the main simulation.
Anyway, it would be best it it can be made to pass. And also it might be a warning of something that can crash also in other contexts.

kthyng:
@knutfrode Ah I don't know why that is coming up now and not before (the small fix I made shouldn't have impacted that?), but I had to fix an issue like that in a test that ran for only vertical_mixing. I'll try to go find it. The same fix from that test should fix the issue in the example, but how should I run the examples locally?

knutfrode:
You can simply run the examples as standalone scripts

knutfrode:
The examples sometimes (as now) detects things that are not covered by the tests

kthyng:
No this was previously caught by a test too, not sure why the example didn't catch it before. Anyway I didn't run this locally but I applied the same fix from a test so hopefully it works.

knutfrode:
Now the same example failed with
```
    ../../examples/example_vertical_mixing.py failed leaving traceback:

      File "/root/project/examples/example_vertical_mixing.py", line 41, in <module>
        o.environment = np.array(np.ones(N, 0)*sea_floor_depth,
                                 ^^^^^^^^^^^^^
      File "/opt/conda/envs/opendrift/lib/python3.11/site-packages/numpy/core/numeric.py", line 191, in ones
        a = empty(shape, dtype, order)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^
    TypeError: Cannot interpret '0' as a data type
```

knutfrode:
The last commit seems to fix the problem (just the HYCOM thredds server hanging in some of the example scripts).
Thus merging now.

PR #1268, opened 2024-04-04
Update variables.py

simonweppe:
Needed otherwise it will fail in `__calculate_derived_environment_variables__()` because 3 inputs are provided i.e.

`em['method'](env, em['input'], em['output'])`


knutfrode:
Yes, you are right.
This was not detected, as this method is deactivated by default, as interpreting negative ocean depths as land might not be correct in all cases.
This "environment mapping mechanism" is a little experimental, and could be modified in the near future, e.g. using the standard config mechanism to activate/deactivate the various mappings. Thus suggestions for improvements are welcome!

PR #1289, opened 2024-04-23
fix z_rho and remove dcrit

kthyng:
Fix z_rho in ROMS reader: z_rho in the ROMS reader could be greater than 0 for wet/dry scenarios when in a dry area. When this happens, z_rho is now changed to nan instead.

Removed dcrit: Previously I had added a parameter to opendrift, dcrit, which was added to the sea surface height as a minimum value in wet/dry cells. However, I have now realized I misinterpreted the ROMS docs on this issue and it should not be present. Therefore, I have removed it from opendrift entirely.


knutfrode:
Seems good.
Is it ready to merge?

kthyng:
@knutfrode Yes, thanks.

PR #1290, opened 2024-04-23
small change to mask values

kthyng:
small change to mask values in `get_lonlats(self)` for values of 9.969209968386869e+36

I am not sure why these values occur but they happen when `coastline_action` is set to "stranding" and `seafloor_action` to "deactivate" and then I can't plot the results. I couldn't find a way to avoid the values in the first place instead.

I'm not sure is this is a universal problem or solution. What do you think @knutfrode?


gauteh:
It almost seems like we incorrectly convert Nan to float somewhere. Do you have a test simulation where this happens regularly?

kthyng:
@knutfrode I updated and yes your new changes fix my problem. Thank you!

PR #1294, opened 2024-04-26
Moving basereader.prepare() to variables.prepare(), as the former was…

knutfrode:
… overriding structured.prepare() due to multiple inheritance, and thus config *drift:max_speed* was not applied if config setting was made after reader was added. Also increasing *drift:max_speed* of OceanDrift from 1 to 2m/s


knutfrode:
@gauteh Til info, se kommentar om multiple inheritance.

PR #1295, opened 2024-05-10
Small changes to FTLE method

mateuszmatu:
Added some small changes to the FTLE code, as described in email to Knut-Frode. The fields look similar, but values now tend to be more positive. This is more apparent for model data than for the double gyre example. 
Not sure if values are 100% correct now (or if they even matter).
I expect that majority of values in a domain of real flow should be positive, as negative FTLE values describe non-chaotic flow.

### DIFF
--- a/opendrift/models/physics_methods.py
+++ b/opendrift/models/physics_methods.py
@@ -515,7 +515,9 @@ def ftle(X, Y, delta, duration):
     # From Johannes Rohrs
     nx = X.shape[0]
     ny = X.shape[1]
-    J = np.empty([nx,ny,2,2], np.float32)
+
+    #Changes by Mateusz M
+    """J = np.empty([nx,ny,2,2], np.float32)
     FTLE = np.empty([nx,ny], np.float32)
 
     # gradient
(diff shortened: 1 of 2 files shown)


opendrift/models/physics_methods.py
```diff
@@ -527,15 +529,36 @@ def ftle(X, Y, delta, duration):
...
             lamda = np.linalg.eigvals(D)
             FTLE[i,j] = np.log(np.sqrt(max(lamda)))/np.abs(duration)
+    """
+    #from np.empty to np.zeros because new method doesn't work on boundaries. Empry left strange values.
+    FTLE = np.zeros([nx,ny])
+    for i in range(1,nx-1):
+        for j in range(1,ny-1):
+            #Centering grid on particle and calculating Jacobian of flow map
+            J = np.zeros([2,2])
+            J[0,0] = (X[i+1, j]-X[i-1, j])/(2*delta)
```
knutfrode, line 547: This looks generally good. But is there a particular reason why you have moved calculation of Jacobian from outside of the loop and into the loop? I fear that this will give much longer computation time, as the overhead of each calculation is now repeated typically up to order of a million times (~if 1000 x 1000).
Did you notice such a slowdown?

mateuszmatu: I see your point and yes, I noticed that it became somewhat slower. There should be a way to take it out of the loop, but I don't see how right now.

PR #1304, opened 2024-05-27
Modification so Reader.get_variables_interpolated() is always called …

Ugomartinez:
…with named arguments. Ad some more documentation in ops and readerops


gauteh:
Looks good!

PR #1307, opened 2024-05-28
open_mfdataset_overlap is much faster if chunksize is hardcoded to sa…

knutfrode:
…me as original data

### DIFF
--- a/opendrift/readers/__init__.py
+++ b/opendrift/readers/__init__.py
@@ -34,11 +34,12 @@ def open_mfdataset_overlap(url_base, time_series=None, start_time=None, end_time
     urls = [t.strftime(url_base) for t in time_series]
     time_step = time_series[1] - time_series[0]
     print('Opening individual URLs...')
-    datasets = [xr.open_dataset(u, chunks='auto').sel({timedim: slice(t, t+time_step-timedelta(seconds=1))})
+    chunks = {'time': 1, 'depth': 1, 'Y': 17, 'X': 2602}
+    datasets = [xr.open_dataset(u, chunks=chunks).sel({timedim: slice(t, t+time_step-timedelta(seconds=1))})
                 for u,t in zip(urls, time_series)]
     print('Concatenating...')
     ds = xr.concat(datasets, dim=timedim,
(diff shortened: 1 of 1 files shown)


opendrift/readers/__init__.py
```diff
@@ -34,11 +34,12 @@ def open_mfdataset_overlap(url_base, time_series=None, start_time=None, end_time
     urls = [t.strftime(url_base) for t in time_series]
     time_step = time_series[1] - time_series[0]
     print('Opening individual URLs...')
-    datasets = [xr.open_dataset(u, chunks='auto').sel({timedim: slice(t, t+time_step-timedelta(seconds=1))})
+    chunks = {'time': 1, 'depth': 1, 'Y': 17, 'X': 2602}
```
knutfrode, line 37: @gauteh The solution to the performance problem seems to be to specify/hardcode that chunksizes should be equal to the original OPeNDAP dataset (Here NorKyst on Thredds).
Do you see a dynamic way to achieve this?
Both `None` or `auto` give extremely slow performance.

knutfrode: After hardcoding chunksizes to be as in original files, manual aggregate is now "only" 5 times slower than thredds aggregate:
https://opendrift.github.io/gallery/example_manual_aggregate.html#sphx-glr-gallery-example-manual-aggregate-py

I will try this approach to set chunksizes dynamically
https://github.com/pydata/xarray/issues/1440#issuecomment-1470801895
On that page there is also a suggestion for specifying `chunks="disk"`, but unfortunately not implemented.

gauteh: Sounds good! Will `chunks=disk` work over opendap? Somehow the chunk size must be available in the metadata. Did you test on files on disk or opendap files?

knutfrode: I don't think `chunks=disk` is implemented.
But yes, in our case I guess we can only align chunks with what is returned from OPeNDAP. And I guess this is the optimal alignment anyways? Unlike many other applications, we want as small chunks as possible, to read a minimum of data.

PR #1313, opened 2024-06-11
Extract water from surface to the draft for icebergs

AchrefAO:
Extract water from surface to the draft for icebergs (to depth+1 only )


knutfrode:
It seems something went wrong with this PR, as it only contains a duplicate of reader_netCDF_CF_generic, and no updates to openberg

PR #1319, opened 2024-06-17
Magnesim/radionuclide

magnesim:


knutfrode:
One test is failing, but this is perhaps work in progress?
`FAILED tests/radionuclide/test_model.py::test_radio_nuclide - ValueError: No config setting named radionuclide:transfer_setup`

PR #1340, opened 2024-06-27
New coastline interaction functionality implemented

TheSylex:
Now you can specify a precision value in degrees to which the particles get their position approximated to the coastline whenever they get stranded.
So we're going from this:
old_stranded - 25s
To this:
0 001deg - 26s
At barely any cost! 😁


knutfrode:
Looks very good!
Merging it now since all tests are passing, but will play with the feature later.

knutfrode:
For the reference:
This feature is now included in OpenDrift 1.11.11, and it switched on by default, with accuracy of 0.001 degrees (~100 meters).
A basic illustration is added in this example:
https://opendrift.github.io/gallery/example_coastline_options.html

PR #1353, opened 2024-07-04
I1352 openberg

AchrefAO:
A corrected version of the icebergs module


knutfrode:
It seems there is some conflicts here, so that it cannot be automatically merged.There are edits to example_openberg_new.py, but this was renamed to example_openberg.py some time ago.

PR #1362, opened 2024-07-15
Addinf the possibility to combine many readers all at once

Ugomartinez:
General modification: modify reader_ops to add the possibility for reader_b to be a list of readers.
Gaussian changes : modify reader_ops and ops to change the operator function defined in ops from gaussian melting to only gaussian factor. The end of the calculation in done by hand, depending on the case.

Maybe we should keep only one representation now, by inserting a reader in a list if it's the only one that is given during the definition. That would allow us to keep only the branches with the lists in the code, making it clearer.
It is also needed to make the code about the profiles combination in reader_ops, I was not understanding what is needed here. I think it is strange to apply any operation on profiles.

### DIFF
--- a/opendrift/readers/operators/ops.py
+++ b/opendrift/readers/operators/ops.py
@@ -46,18 +46,43 @@ class Combine:
         """
         from .readerops import Combined as ReaderCombined
 
-        def gaussian_melting(a, b, lon, lat, lon_center, lat_center, std):
+        def gaussian_factor(lon, lat, lon_center, lat_center, std):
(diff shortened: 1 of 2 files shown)


knutfrode:
Looks good to me!
Ready to merge?

Ugomartinez:
I think it is 👍

Ugo

Le 15 juil. 2024 à 14:39, Knut-Frode Dagestad ***@***.***> a écrit :

﻿

Looks good to me!
Ready to merge?

PR #1377, opened 2024-08-21
Quick implementation of progress bar option for the run method

nordam:


gauteh:
This is pretty useful, but remember that this may cause trouble if you are not attached to a tty: https://github.com/gauteh/lieer/blob/eb5ff8a072d7bbbe977049f384348023123d1892/lieer/gmailieer.py#L471

nordam:
Ah, didn't think of that. Probably good to implement some checks.

Issue #1379, opened 2024-08-22
issue with Reader roms native is discarded after failing more times than allowed


Wen77777777:
When using ROMS model data for a test, the error ERROR opendrift.models.basemodel.environment:666: list index out of range occurs

import numpy as np
from opendrift.readers import reader_ROMS_native
from opendrift.models.oceandrift import OceanDrift

o = OceanDrift(loglevel=20)

roms_file = "E:\his_6238.nc"
reader_roms = reader_ROMS_native.Reader(roms_file)
o.add_reader(reader_roms)

o.seed_elements(lon=119.4, lat=37.8, radius=0, number=10,
                z=np.linspace(0, -10, 10), time=reader_roms.start_time)

o.run(time_step=3600)
############################
ERROR   opendrift.models.basemodel.environment:666: list index out of range
Traceback (most recent call last):
  File "h:\AnacondaEnvs\opendrift\Lib\site-packages\opendrift\models\basemodel\environment.py", line 641, in get_environment
    reader.get_variables_interpolated(
  File "h:\AnacondaEnvs\opendrift\Lib\site-packages\opendrift\readers\basereader\variables.py", line 904, in get_variables_interpolated
    env, env_profiles = self.get_variables_interpolated_xy(
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "h:\AnacondaEnvs\opendrift\Lib\site-packages\opendrift\readers\basereader\variables.py", line 770, in get_variables_interpolated_xy
    env, env_profiles = self._get_variables_interpolated_(
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "h:\AnacondaEnvs\opendrift\Lib\site-packages\opendrift\readers\basereader\structured.py", line 281, in _get_variables_interpolated_
    self.get_variables(blockvariables_before, time_before,
  File "h:\AnacondaEnvs\opendrift\Lib\site-packages\opendrift\readers\reader_ROMS_native.py", line 538, in get_variables
    var = self.Dataset.variables[varname[0]]
                                 ~~~~~~~^^^
IndexError: list index out of range
09:17:39 INFO    opendrift.models.basemodel.environment:668: ========================
09:17:39 WARNING opendrift.models.basemodel.environment:673: Reader roms native is discarded after failing more times than allowed (1)
##################################

knutfrode:
Hi, can you make available this ROMS file for testing?

[...]

PR #1382, opened 2024-08-23
Cleaned mapping of variables in ROMS native reader. If multiple varia…

knutfrode:
…bles mapping to same standard_name, a provided mapping takes precedence. east/north-variables are now mapped to x/y-standard_names, since ROMS reader always uses lonlat CRS (east=x, north=y), and takes predominance over x-y-variables, as the former do not need to be rotated.


knutfrode:
@kthyng FYI, please not these updates in ROMS native reader, where I also updated your ROMS tests accordingly.

The ROMS reader is still quite messy, so I will probably do some more cleaning in the near future. But at least this PR solves some other issues, e.g. https://github.com/OpenDrift/opendrift/issues/1379

PR #1470, opened 2024-12-30
Replaced mambaforge with miniforge in circleCI config

knutfrode:


knutfrode:
I wonder if failing examples are due to Python 3.12 and missing netCDF depedency

PR #1471, opened 2025-01-03
[run-ex] set_up_map now stores crs_plot and crs_lonlat for later user…

knutfrode:
…. Landmask is now obtained on Mercator/plot grid (and not lonlat-grid) when fast=True


knutfrode:
@gauteh 
This is the difference (from pytest `test_fast_norkyst`) in fast-landmask after this change, i.e. obtaining landmaks on a regular grid in Mercator coordinates, instead of lonlats.
I.e. basically the same, but pixels with rounded edges.
This is significantly faster for some cases, e.g. with extreme aspect ratio with particles at 178W, 60N and 180E, 60N, for which case the original is very slow. 

result
baseline

knutfrode:
I guess this is due to roaring landmask stored natively on a lonlat-grid. 
Thus storing on a Mercator grid might be better for plotting, and would give square land-pixels on plots.
However, during simulations, positions/particles are checked as lonlat-values, and then it is probably faster with roaring landmask on lonlat grid, as presently.

gauteh:
Nice! I don't quite understand why you get rounded pixels.. maybe numerical rounding?  Anyway, the rounding is probably not problem.

knutfrode:
@gauteh 
This seem to have a small positive impact (faster) for general cases with fast/raster landmask, and a large positive impact in rare cases with unnatural aspect ratio (e.g. 175W to 175E and 60-61N).
Also not completely sure why pixels get rounded, but they seem correctly located.

To merge this, and let the graphical test to pass, shall I simply replace the old `test_fast_norkyst.png` with the new one (with rounded pixels)?

PR #1476, opened 2025-01-13
[run-ex] Using RoaringLandmask to check/screen if there is any land w…

knutfrode:
…ithin plot area, otherwise skip plotting of vector coastline.


knutfrode:
@gauteh This screening gives a large speedup for the rather common case that there is no (raster) land pixels within a plot area.

PR #1486, opened 2025-01-27
[run-ex] Preparing for storing output in Xarray Dataset (self.result)…

knutfrode:
… instead of self.history


knutfrode:
Failing due to maintainance at thredds.met.no

PR #1491, opened 2025-02-03
[run-ex] Major change: simulation result is stored in Xarray Dataset self.result, replacing recarray self.history

knutfrode:


examples/example_compare_oilbudgets.py
```diff
@@ -36,11 +36,11 @@
         o.run(duration=timedelta(hours=hours), time_step=600)
         b[ot] = o.get_oil_budget()
         # Get viscosity and density
-        kin_viscosity = o.history['viscosity']
-        dyn_viscosity = kin_viscosity * o.history['density'] * 1000 # mPas
-        viscosities[ot] = dyn_viscosity.mean(axis=0)
-        densities[ot] = o.history['density'].mean(axis=0)
-        water_fractions[ot] = o.history['water_fraction'].mean(axis=0)
+        kin_viscosity = o.result.viscosity
```
gauteh, line 39: I think this looks clearer now!

opendrift/export/io_netcdf.py
```diff
@@ -3,235 +3,82 @@
...
-# Strives to be compliant with netCDF CF-convention on trajectories
-# https://cfconventions.org/Data/cf-conventions/cf-conventions-1.6/build/cf-conventions.html#idp8377728
-# https://geo-ide.noaa.gov/wiki/index.php?title=NODC_NetCDF_Trajectory_Template
-
-skip_parameters = ['ID']  # Do not write to file
+# Module with functions to export/import trajectory data to/from netCDF file.
+# Follows netCDF CF-convention on trajectories:
+# https://cfconventions.org/Data/cf-conventions/cf-conventions-1.12/cf-conventions.html#_multidimensional_array_representation_of_trajectories
 
+def datetime_from_datetime64(dt64):
```
gauteh, line 18: May get some help from pd.to_datetime here.

opendrift/export/io_netcdf.py
```diff
@@ -3,235 +3,82 @@
...
-        var = self.outfile.createVariable(prop, dtype, ('trajectory', 'time'), fill_value=fill)
-        for subprop in self.history_metadata[prop].items():
-            if subprop[0] not in ['dtype', 'constant', 'default', 'seed']:
-                # Apparently axis attribute shall not be given for lon and lat:
-                if prop in ['lon', 'lat'] and subprop[0] == 'axis':
-                    continue
-                var.setncattr(subprop[0], subprop[1])
+    if os.path.exists(self.outfile_name):
+        logger.warning(f'Deleting existing {self.outfile_name}')
+        os.remove(self.outfile_name)
```
gauteh, line 28: Better to raise an error? Unless user supplies a force/overwrite flag?

opendrift/export/io_netcdf.py
gauteh: One thing I have run into many times with xarray and netcdf + thredds is that xarray uses nano-seconds unless forced to do otherwise. As well as variable offset, which means that concatenating files using thredds aggregates do not work unless they have the same offset (reference for time). This is maybe not a problem here.

opendrift/export/io_netcdf.py
```diff
@@ -3,235 +3,82 @@
...
-
-            for att in src.ncattrs():  # Copy global attributes
-                dst.setncattr(att, src.getncattr(att))
-
-        move(self.outfile_name + '_tmp', self.outfile_name)  # Replace original
-    except Exception as me:
-        print(me)
-        print('Could not convert netCDF file from unlimited to fixed dimension. Could be due to netCDF library incompatibility(?)')
+    logger.debug('Making netCDF file CDM compliant with fixed dimensions')
+    self.result.to_netcdf(self.outfile_name + '_tmp', unlimited_dims={})
```
gauteh, line 79: What if _tmp exists. Better to use python tempfile.

opendrift/models/basemodel/__init__.py
```diff
@@ -2286,86 +2243,95 @@ def state_to_xarray(self):
...
+                            'constant_reader') and var in self.env.readers[
+                                readers[0]]._parameter_value_map:
+                        self.add_metadata(
+                            keyword, self.env.readers[
+                                readers[0]]._parameter_value_map[var][0])
+                    else:
+                        self.add_metadata(keyword, self.env.priority_list[var])
+            self.metadata_dict = {key:str(value) for key,value in self.metadata_dict.items()}
+
+            final_metadata = {
```
gauteh, line 2299: Can maybe use trajan to generate this?

opendrift/export/io_netcdf.py
knutfrode: ok, yes that sounds bad.
Presently there are some ad_hoc pd.to_datetime and methods like the above scattered around.

But we should probably be consistent on that all input datetimes are immediately converted to some internal standard (e.g. with pd.to_datetime).

opendrift/export/io_netcdf.py
knutfrode: Overwrite was always the default. The above might be unnecessary, basically a debug comment to show that initialisation of file is done.

Before, io_init() created the netCDF-file with dimensions and attributes etc, but no nothing else then the above happens at init, and the first write is a single call to `self.result.to_netcdf()`
Thus letting Xarray do the job of laying out the netCDF file structure.

opendrift/export/io_netcdf.py
knutfrode: Yes, will do

opendrift/export/io_netcdf.py
gauteh: Ok. I guess this is how netcdf works as well.

gauteh:
Impressive change, but looks like overall less and cleaner code than before.

knutfrode:
> Impressive change, but looks like overall less and cleaner code than before.

There are still many legacy things that can be removed or simplified even further.
Will now first make all tests pass, and thereafter continue cleanup.

gauteh:
> > Impressive change, but looks like overall less and cleaner code than before.
> 
> There are still many legacy things that can be removed or simplified even further. Will now first make all tests pass, and thereafter continue cleanup.

And eventually we will also be able to use the result directly in Trajan, which is another win.

PR #1504, opened 2025-02-10
[draft] improved closest ocean points for shape readers

poplarShift:
Hi, I've had a round with the detection of closest ocean points. For shape readers, this is pretty easy, see this PR. I assume there are some things to fix but wanted to hear your ideas first. I am also unsure about using a global config argument for the buffer, and how to name it.

Also attaching a before/after plot where the current PR keeps particles from "tunneling" through narrow pieces of land. @mateuszmatu might have some more detailed plots of what the current algorithm's helper grid looks like for such cases.
 
Before:
13225-old
After:
13225-new

### DIFF
--- a/opendrift/models/basemodel/__init__.py
+++ b/opendrift/models/basemodel/__init__.py
@@ -329,6 +329,22 @@ class OpenDriftSimulation(PhysicsMethods, Timeable, Configurable):
                 'description': 'The precision of the particle position approximation to the coastline.',
                 'level': CONFIG_LEVEL_BASIC
             },
+            'general:land_to_ocean_move_buffer': {
+                'type':
+                'float',
+                'min':
+                0.0,
+                'max':
+                np.inf,
+                'default':
+                0,
+                'units':
+                'shape reader projection units',
+                'level':
+                CONFIG_LEVEL_ADVANCED,
+                'description':
+                'Move land particles this many projection units into ocean.'
+            },            
             'general:time_step_minutes': {
                 'type': 'float',
                 'min': .01,
@@ -891,14 +907,8 @@ class OpenDriftSimulation(PhysicsMethods, Timeable, Configurable):
 
     def closest_ocean_points(self, lon, lat):
         """Return the closest ocean points for given lon, lat"""
+        from opendrift.readers.reader_shape import Reader as ShapeReader
 
-        deltalon = 0.01  # grid
-        deltalat = 0.01
-        numbuffer = 10
-        lonmin = lon.min() - deltalon * numbuffer
-        lonmax = lon.max() + deltalon * numbuffer
-        latmin = lat.min() - deltalat * numbuffer
-        latmax = lat.max() + deltalat * numbuffer
         if not 'land_binary_mask' in self.env.priority_list:
             logger.info('No land reader added, '
                         'making a temporary landmask reader')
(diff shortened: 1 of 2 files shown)


knutfrode:
Hi, yes this looks great for your example.

However, as you mention, there are some concerns:

- config key `general:land_to_ocean_move_buffer` does not sound very good/intuitive, so we should consider a better name, or whether this is strictly necessary at all.

- This needs GeoPandas, which is a great library - however, we should limit the number of external requirements to an absolute minimum. Is it possible to achieve the same with the libraries already available? Alternatively, could GeoPandas be made an optional dependency, so that tests will pass also without GeoPandas installed?

opendrift/models/basemodel/__init__.py
```diff
@@ -329,6 +329,22 @@ def __init__(self,
                 'description': 'The precision of the particle position approximation to the coastline.',
                 'level': CONFIG_LEVEL_BASIC
             },
+            'general:land_to_ocean_move_buffer': {
+                'type':
+                'float',
```
knutfrode, line 334: `key: value` on the same line would be much more readable here

poplarShift:
> Hi, yes this looks great for your example.
> 
> However, as you mention, there are some concerns:
> 
>     * config key `general:land_to_ocean_move_buffer` does not sound very good/intuitive, so we should consider a better name, or whether this is strictly necessary at all.
> 
>     * This needs GeoPandas, which is a great library - however, we should limit the number of external requirements to an absolute minimum. Is it possible to achieve the same with the libraries already available? Alternatively, could GeoPandas be made an optional dependency, so that tests will pass also without GeoPandas installed?

Well, I saw opendrift was already using geopandas. I would have to look up if geopandas does something more fancy/faster than just python looping over all geometries and doing a point in polygon test.
```
$ git blame opendrift/models/basemodel/__init__.py|grep geopandas

658f841d7 opendrift/models/basemodel/__init__.py (Knut-Frode Dagestad    2023-11-21 09:57:25 +0100 2773)                 import geopandas as gpd
```

knutfrode:
Yes, GeoPandas is presently used in one rare use case, but not needed with general usage and by any of the automated tests.
Thus if you are able to use any existing dependencies instead, that would be the best.
If difficult to circumvent, I suggest that we can add a check so that this algorithm is used only if GeoPandas would be available (with a warning/hint that installing GeoPandas would activate this feature). At a later stage, GeoPandas might become a regular dependency anyway, and the check can be removed.

Regarding the buffer, can it be hardcoded as 0 to begin with?
Several downstream (web) clients of OpenDrift (e.g. OpenAPI) do not like config settings to disappear or being renamed, as it forces version updates and rebuild of frontends. Thus we should think thrice before adding something.

An alternative to using the config mechanism, could be to have an object property (e.g. `self._land_to_ocean_move_buffer`) that could be adjusted by the user without touching the config mechanism. I think this is not very healthy practice in the long run, but it could be a temporary solution.

poplarShift:
I've updated the PR without geopandas and argument to the Shape reader . Note I now use the awkwardly named `__on_land__` method outside the class.

Can be tested with this code snippet - let me know where to put a test.
```
from opendrift.readers.reader_shape import Reader
import numpy as np

self = Reader.from_shpfiles(...)

x = np.random.random_sample((10_000, ))*5+5
y = np.random.random_sample((10_000, ))*3+62

self.get_nearest_outside(x, y, 0)
``

knutfrode:
Yes, I think this looks good, and I can merge it.

A test would be good, and can e.g. be placed here:
https://github.com/OpenDrift/opendrift/blob/master/tests/models/test_seed.py

But one question:
You have made `land_buffer_distance` and input parameter to the constructor `from_shpfiles`
But it sounds more logically to have this (only) as a parameter for the actual call to `get_nearest_outside`, since in principle (later one day) one could perform repeated calls to the same shapefile with different buffer sizes.
The reason for your implementation is perhaps because you want to control the buffer size at creation time, to avoid having it as a dedicated config setting?

PR #1519, opened 2025-02-20
formating and remove poetry

gauteh:
- yapf: don't allow splitting dict values
- yapf: format basemodel


gauteh:
I don't think these errors are related. Probably netCDF version issues.

PR #1553, opened 2025-03-19
clarify log info for default landmask

poplarShift:
To me, the current log info for the default landmask is very confusing. The `reader_global_landmask` module does not even mention `config:drift_max_speed`.


gauteh:
Yes, that was left-over from the old implementation.

PR #1555, opened 2025-03-24
Handle uninstalled dependencies when reporting their versions

poplarShift:
Currently, openruns crash when software is not installed whose version `versions` is trying to report. This seems excessive.

### DIFF
--- a/opendrift/__init__.py
+++ b/opendrift/__init__.py
@@ -82,7 +82,7 @@ def versions():
     def version(package):
         try:
             return importlib.metadata.version(package)
-        except:
+        except ImportError:
             return "N/A"
         
     import multiprocessing


opendrift/__init__.py
```diff
@@ -79,7 +79,12 @@ def open(filename):
     return o
 
 def versions():
-    from importlib.metadata import version
+    def version(package):
+        try:
+            return importlib.metadata.version(package)
+        except:
```
gauteh, line 85: Can you filter on a specific exception here? What is the reason that you run into this. Normally this works when either in a git repository or installed setting.

gauteh: So the reason is that `importlib` isn't available?

poplarShift: No, `adios_db`

poplarShift:
Hi @knutfrode there was a mistake in which error is filtered on - `version` throws not an `ImportError` but `importlib.metadata.PackageNotFoundError`

knutfrode:
Ok, I updated directly on master
https://github.com/OpenDrift/opendrift/commit/cad44d6acf05aa8f026a5fbe78f3d8e302361c7d

PR #1569, opened 2025-04-11
ChemD: Saving custom attrs to self.result and refer to these in postprocessing

manuelaghito:
can later be moved to post_run()
ready to merge


manuelaghito:
needed to add an additional check for multidimentional arrays. For example self.result['ntransformations']=self.ntransformations assignment did not work because self.ntransformation is 2D

PR #1593, opened 2025-05-09
ChemD: removing old temporary fix for 0 values in shyfem datasets

manuelaghito:
removing old fix
this was due to erroneous 0 values in some old shyfem datasets
better to change/filter the input datasets if the problem still exists
ready to merge


manuelaghito:
@loriscalgaro let me know if this still causes problems

PR #1600, opened 2025-05-14
Optimized trajectory indexing in import_file

manuelaghito:
faster loading of large opendrift output files. In the earlier version, manual indexing with np.arange(num_elements) and index_of_last would trigger a complete materialization of the self.result[var]. This version allows lazy loading with Dask. 

AI recommended. Human check welcome. Potential drawbacks?

### DIFF
--- a/opendrift/export/io_netcdf.py
+++ b/opendrift/export/io_netcdf.py
@@ -117,7 +117,15 @@ def import_file(self, filename):
     kwargs = {}
     for var in self.result.data_vars:
         if var in self.ElementType.variables:
-            kwargs[var] = self.result[var][np.arange(num_elements), index_of_last]
+            last_vals_var = xr.apply_ufunc(
+                lambda arr, i: arr[i],
+                self.result[var], index_of_last,
+                input_core_dims=[['time'], []],
+                output_core_dims=[[]],
+                vectorize=True, dask='parallelized',
+                output_dtypes=[self.result[var].dtype]
+            )
+            kwargs[var] = last_vals_var.compute()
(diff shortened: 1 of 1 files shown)


manuelaghito:
you can also test this for example with /lustre/storeA/users/manuel/produsertvann_Snorre_WCM2024/Output/20250512-141015-ChemicalDrift_producedwater_mp_2024_v3-Acenaphthene-6h-sim_365d-em_365d/concatenated.nc

knutfrode:
I tested locally, and confirm it works much faster (`o = opendrift.open('large_file.nc')`).
That said, I guess the time saved upon initialisation is lost later when accessing data from self.result. But at least we don't waste time on importing the things that are never used - which is good.
Not totally sure how this improvement work, but since all tests are passing, I assume it is all safe.
So merging now.

manuelaghito:
I guess it is better in most cases because when you reload the file often you are not going to access all the data.

possibly .persist() is better than .compute() in many situations, if the same data is later accessed many times.

.compute() Triggers immediate computation and returns a concrete (non-Dask) object.
.persist() Triggers computation but returns a Dask object with results stored in memory.

knutfrode:
We could in principle add possibility to chose:
`o = opendrift.open('large_file.nc', method="persist")`

Or perhaps a better way for the future is to allow the user to open the dataset with e.g. Xarray, and then create an OpenDrift-object from this:
```
ds = xarray.open_dataset('large_file.nc', <xarray options>)
o = opendrift.from_xarray(ds)
```

PR #1749, opened 2025-12-11
Fix parquet io

kthyng:
Looks like the parquet exporter wasn't updated when output changes happened with OpenDrift. This is my attempt to fix it. As a disclaimer, I wrote these changes using ChatGPT because I was not familiar enough with parquet nor the details of the original code to get it right. I tested it with OpenDrift's tests and my own subsidiary package tests and it looks like it is working but I encourage an extra eye on this if possible.

### DIFF
--- a/opendrift/models/basemodel/__init__.py
+++ b/opendrift/models/basemodel/__init__.py
@@ -2240,17 +2240,6 @@ class OpenDriftSimulation(PhysicsMethods, Timeable, Configurable):
 
         return self.result
 
-        # if export_buffer_length is None:
-        #     pass  # TODO - do this for self.result
-        # else:  # If output has been flushed to file during run, we
-        #     # need to reimport from file to get all data in memory
-        #     del self.environment
-        #     if hasattr(self, 'environment_profiles'):
-        #         del self.environment_profiles
-        #     self.result = xr.open_dataset(outfile)
-
-        # return self.result
-
     def increase_age_and_retire(self):
         """Increase age of elements, and retire if older than config setting."""
         # Increase age of elements


opendrift/models/basemodel/__init__.py
```diff
@@ -2225,14 +2225,18 @@ def run(self,
...
+            if ext in [".nc", ".nc4", ".netcdf"]:
+                # NetCDF: open with xarray as before
+                self.result = xr.open_dataset(outfile)
+            elif ext == ".parquet":
+                # Parquet: let the parquet exporter set self.result in its close()
+                # (we do nothing here)
+                pass
+            else:
+                # Fallback: try normal xarray open
+                self.result = xr.open_dataset(outfile)
```
knutfrode: The lines 2230-2239 also seem a bit overly complicated. Could this not be simply:
```
if outfile is not None:
    ext = os.path.splitext(outfile)[1].lower()
    if ext == ".parquet":
         pass  # let the parquet exporter set self.result in its close()
    else:
         # reimport from file to get the complete simulation (lazily) in memory
         self.result = xr.open_dataset(outfile)  
```

opendrift/models/basemodel/__init__.py
```diff
@@ -2225,14 +2225,18 @@ def run(self,
         # Remove any elements scheduled for deactivation during last step
         self.remove_deactivated_elements()
 
-        if export_buffer_length is None:
-            pass  # TODO - do this for self.result
-        else:  # If output has been flushed to file during run, we
-            # need to reimport from file to get all data in memory
-            del self.environment
-            if hasattr(self, 'environment_profiles'):
-                del self.environment_profiles
```
knutfrode: The deletion of `self.environment` and `self.environment_profiles` has here been removed. Tests are passing, so might be ok, but I seem to recall that this deletion was necessary for some reason. Was this removed deliberately, or was it just ChatGPT not seing the point?

opendrift/models/basemodel/__init__.py
knutfrode: Also I am note completely sure about replacing `export_buffer_length is None` with `outfile is None`,
as one could have an outfile but still set `export_buffer_length` to `None` to get just a single write at the end.
In this case it would be unnecessary to reimport from that file, as the complete simulation is already in memory.
On the other hand, it might perhaps be ok for consistency.

knutfrode:
Hi,
Apart from the small concerns regarding the updates to `basemodel` (see comments above), this seems fine to me.

However, I have never used the parquet writer, so perhaps @poplarShift or @mateuszmatu would be interested to have a look before merging this?
If no comments until tomorrow, I guess we can just merge it, and they can have a look later.
The writer is anyway not functional at the moment.

kthyng:
@knutfrode Thanks for your comments. I can integrate them later today or Monday in case we hear from @poplarShift first.

PR #1752, opened 2025-12-19
Two FVCOM readers and a very simple salmon louse module

poplarShift:
Here come three contributions from Akvaplan-niva.
- An xarray-based FVCOM reader (can read anything xarray can read), which gives access to e.g. xarray's caching routines for free. It currently only looks for nearest neighbours, no interpolation done.
- A netCDF based FVCOM reader which @HvardE may say more about (it reads time steps ahead)
- A very simple salmon louse model with vertical behaviour.

### DIFF
--- a/opendrift/models/basemodel/__init__.py
+++ b/opendrift/models/basemodel/__init__.py
@@ -2232,6 +2232,9 @@ class OpenDriftSimulation(PhysicsMethods, Timeable, Configurable):
             del self.environment
             if hasattr(self, 'environment_profiles'):
                 del self.environment_profiles
+            self.result = xr.open_dataset(outfile)
+
+        return self.result
 
     def increase_age_and_retire(self):
         """Increase age of elements, and retire if older than config setting."""


knutfrode:
Great, very interesting!

Tests are failing. It seems that pykdtree is missing. It could perhaps be optional, but in the meantime to get through the tests, you may add pykdtree to environment.yml.
But it also seems that some finite volume interpolation routine is missing?
We try to be a little restrictive with the amount of external dependencies, so regarding numba, it might be an option that it is not installed by default, but rather that the reader will state clearly that numba must be installed for it to work. (Normal exception on missing import might be enough, though)

We try to keep one topic per PR to facilitate getting a better overview, so you can consider that for next time.

The asynchronous reading of next time step has been planned for the generic netCDF reader for some time, so interesting to see that this works well for your case (I presume?).

In principle there should not be a need for a dedicated FVCOM reader, and the one presently contained in OpenDrift was renamed to reader_netCDF_CF_unstructured.py
Output from unstructured models are all supposed to follow the UGRID conventions, but unfortunately the compliance is not very good - and I believe FVCOM is one of the least compliant unstructured models.
There is a plan to further develop the generic-unstructured reader to be based on the XUgrid package: https://deltares.github.io/xugrid/
and then this should ideally be able to read FVCOM data - by adding some model specific codelines to make it compliant with XUgrid and UGRID conventions.
Did you try XUgrid yourselves?

PR #1755, opened 2026-01-06
Adds a FVCOM reader which builds on netCDF4. Reads the next timestep …

poplarShift:
…synchronously. In practice, this allows you to lower the model timestep without penalty in terms of simulation time, since you won't need to wait for the next timestep to load as you enter it. (with an appropriately low timestep, that is)

Supersedes #1752

### DIFF
--- a/opendrift/readers/basereader/finite_volume.py
+++ b/opendrift/readers/basereader/finite_volume.py
@@ -1,0 +1,446 @@
+from abc import abstractmethod
+import numpy as np
+import logging
+
+logger = logging.getLogger(__name__)
+
+from .variables import Variables
+from opendrift.readers.interpolation.finite_volume import FiniteVolumeReaderBlock
+from numba import njit, prange
+
+class FiniteVolumeReader(Variables):
+    """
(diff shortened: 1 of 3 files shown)


HvardE:
> > The only reason why it's not in the xarray reader is that we've worked on the readers separately. I would also like to add the vertical interpolation scheme to the xarray reader at some point.
> 
> Ok, that is understandable. But then a common goal is that these two FVCOM readers should be merged in the future - and with the ultimate goal of a generic unstructured reader that works for all unstructured models.

Yep, and I think we can do that within a few months - so to keep things clearer, I think don't need to merge this reader into the repository, but rather update the xarray one with the methods proposed in this reader (y)

knutfrode:
Yes, I agree it is better to not load too much "work in progess" into the main repository.
But then we can let this PR stay unmerged, still allowing others to peek into the stuff to come.

Btw, for testing it would be useful to have some "standard" or "reference" FVCOM output files to test on. But I am not ware of any such, and understand the different versions of FVCOM can have slightly different output.
If you are aware of any such datasets, we could make some FVCOM-specific unit-tests to ensure the reader will work with all variants.

PR #1798, opened 2026-04-13
Update reader_ROMS_native.py

Erispoe29:
Hello,

I have started using the Opendrift library, which is incredibly easy to get to grips with.
 am using only the surface fields from Croco’s output fields. This triggered an error message, as with only two spatial dimensions, the reader reader_ROMS_native did not recognise the ‘hc’ variable.  This line "if z.min() == 0 or self.hc is None:" was failing and then the code was not reading x_sea_water_velocity and y_sea_water_velocity. I've added only_surface_fields boolean to bypass the test on "hc" if there are only surface fields.

### DIFF
--- a/opendrift/readers/reader_ROMS_native.py
+++ b/opendrift/readers/reader_ROMS_native.py
@@ -89,6 +89,7 @@ class Reader(BaseReader, StructuredReader):
         self.sea_floor_depth_below_sea_level = None
         self.z_rho_tot = None
         self.s2z_A = None
+		self.only_surface_fields = False
 
         if filename is None:
             raise ValueError('Need filename as argument to constructor')
@@ -190,6 +191,7 @@ class Reader(BaseReader, StructuredReader):
 
         if 's_rho' not in self.Dataset.variables:
             dimensions = 2
+			self.only_surface_fields = True
         else:
             dimensions = 3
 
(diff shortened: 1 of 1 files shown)


knutfrode:
Hi,
If `hc` is missing is should be set no None, and if z is missing it should be set to array 0.
Thus it is strange that this line is failing.
Can you share such a sample file so that I can test?
It would be best if we can avoid complicating with new flags.

Erispoe29:
Hi,
Thank you for your answer. Please find attached a subset of the Croco output I'm using, the associated grid file and the python script to launch all this.
I also tried to add hc variable in the grid file but it led to the same error message.

[subset_opendrift.py](https://github.com/user-attachments/files/26774782/subset_opendrift.py)
[croco_surf_subset.zip](https://github.com/user-attachments/files/26774751/croco_surf_subset.zip)
[croco_grd_subset.zip](https://github.com/user-attachments/files/26774733/croco_grd_subset.zip)

knutfrode:
Yes, it turns out the hc parameter was not set since this was a 2D file.

But your example now works for me after this update:
https://github.com/OpenDrift/opendrift/commit/0325f16caf654ff74a29e9bf45dd8855ca3ea579

Thus closing this PR.

Erispoe29:
Thank you very much for your review and solving this.

